PR #37, opened 2016-11-22
Add the Symfony3 base image

sroze:


symfony/3-apache/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM quay.io/continuouspipe/php-apache:7.0_v1
+
+RUN apt-get update -qq \
+ && DEBIAN_FRONTEND=noninteractive apt-get -yq install \
+    acl \
```
sroze: Is it to have `setfacl` ? If yes, wouldn't in makes sense to move it to the Ubuntu base image actually, as we'll have this problem often I believe!

symfony/3-apache/usr/local/share/symfony/install.sh
```diff
@@ -0,0 +1,29 @@
...
+mkdir -p /app/var
+
+set +e
+is_nfs
+IS_NFS=$?
+set -e
+
+if [ "$IS_NFS" -ne 0 ]; then
+  # Ensure code is owned by a user other than the web server user
+  chown -R build:build /app
```
sroze: That's quite aggresive

symfony/3-apache/usr/local/share/symfony/install.sh
```diff
@@ -0,0 +1,29 @@
...
+  chmod -R go-rw /app/var
+  # Fix permissions so the web server user can write to /app/var for symfony cache files
+  setfacl -R -m u:www-data:rwX -m u:build:rwX /app/var
```
sroze: `/etc/exports`:
```
/Users/sroze 192.168.99.100 -alldirs -mapall=502:20
```

symfony/3-apache/usr/local/share/symfony/install.sh
sroze: But fair probably.

kierenevans: Same as deploy user owned code in traditional deployments.

symfony/3-apache/Dockerfile
kierenevans: Will move out :)

sroze:
🎉

PR #43, opened 2016-11-24
Fix problems found by linter

kierenevans:
Based on top of #42, fix problems that they found.

### DIFF
--- a/magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento.sh
+++ b/magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento.sh
@@ -2,12 +2,6 @@
 
 set -xe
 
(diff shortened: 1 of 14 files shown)


magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento.sh
```diff
@@ -2,12 +2,6 @@
 
 set -xe
 
-if [ -L "$0" ] ; then
```
sroze: How come the directory will be guessed then?

sroze: Ah 😛

sroze:
👍 

It looks like yup!

PR #48, opened 2016-11-25
Ability to run commands as the code owner

kierenevans:
Ability to run commands as the user that owns the code. Especially useful when using APP_USER_LOCAL mode to set up a random username that matches the user/group IDs that a mountpoint from a volume is set up with.

### DIFF
--- a/ubuntu/16.04/usr/local/share/bootstrap/update_permissions.sh
+++ b/ubuntu/16.04/usr/local/share/bootstrap/update_permissions.sh
@@ -18,9 +18,14 @@ function update_permissions() {
     read -r owner group owner_id group_id < <(stat -c '%U %G %u %g' "$REFERENCE")
     if [[ "$owner" = UNKNOWN ]]; then
         APP_USER="$(randname)"
+        CODE_OWNER="$APP_USER"
         export APP_USER
+        export CODE_OWNER
         if [[ "$group" = UNKNOWN ]]; then
-            export APP_GROUP="$owner"
+            APP_GROUP="$owner"
+            CODE_GROUP="$APP_GROUP"
+            export APP_GROUP
+            export CODE_GROUP
             addgroup --system --gid "$group_id" "$APP_GROUP"
         fi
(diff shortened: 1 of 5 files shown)


ubuntu/16.04/usr/local/share/bootstrap/update_permissions.sh
```diff
@@ -17,12 +17,21 @@ function update_permissions() {
...
-            export APP_GROUP="$owner"
-            addgroup --system --gid "$group_id" "$APP_GROUP"
+            APP_GROUP="$owner"
+            CODE_GROUP="$APP_GROUP"
+            export APP_GROUP
+            export CODE_GROUP
+            groupadd --system --gid "$group_id" "$APP_GROUP" || groupmod -g "$group_id" "$APP_GROUP"
         fi
-        adduser --system --uid="$owner_id" --gid="$group_id" "$APP_USER"
+        useradd --create-home --system --uid "$owner_id" --gid "$group_id" "$APP_USER" || usermod -u "$owner_id" "$APP_USER"
```
andytson-inviqa: I'm wondering if changing the uid of the APP_USER will cause problems because it's home directory will no longer be readable

kierenevans: A very good point that I hadn't considered.

http://manpages.ubuntu.com/manpages/xenial/man8/usermod.8.html says the following:
```
           The user's mailbox, and any files which the user owns and which are
           located in the user's home directory will have the file user ID
           changed automatically.

           The ownership of files outside of the user's home directory must be
           fixed manually.
```

It might be safe to say that everything outside the home directory will be in /app in the case of our image (unless APP_USER is set to a system user).

kierenevans:
Rebased on top of master.

ubuntu/16.04/usr/local/share/bootstrap/update_permissions.sh
andytson-inviqa: yep, sounds safe then

PR #49, opened 2016-11-25
Use variables for code owner

kierenevans:
As the code owner can now be a dynamically named user, always running as the build user would not make sense.

Instead, use the new functionality introduced in #48 to chown files to the new code user and run commands as them as well.

Incorporates #47

### DIFF
--- a/drupal8-apache/7.0/usr/local/share/drupal8/install_finalise.sh
+++ b/drupal8-apache/7.0/usr/local/share/drupal8/install_finalise.sh
@@ -21,13 +21,13 @@ mkdir -p "$SETTINGS_DIR/files/"
 
 # Install a database if there isn't one yet
 set +e
-as_build "drush sql-query 'SHOW TABLES;' | grep -v cache | grep -q ''" /app/docroot
+as_code_owner "drush sql-query 'SHOW TABLES;' | grep -v cache | grep -q ''" /app/docroot
 HAS_CURRENT_TABLES=$?
 set -e
 if [ "$HAS_CURRENT_TABLES" -ne 0 ] && [ -n "$DRUPAL_INSTALL_PROFILE" ]; then
-  chown build:build "$SETTINGS_DIR/files/"
-  as_build "echo 'y' | drush site-install lightning" "/app/docroot"
-  chown -R www-data:www-data "$SETTINGS_DIR/files/"
+  chown "$CODE_OWNER:$CODE_GROUP" "$SETTINGS_DIR/files/"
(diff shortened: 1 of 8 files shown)


kierenevans:
This has been rebased on top of the latest master

drupal8-apache/7.0/usr/local/share/drupal8/install_finalise.sh
```diff
@@ -20,9 +20,14 @@ chmod u+w "$SETTINGS_DIR" || true
 mkdir -p "$SETTINGS_DIR/files/"
 
 # Install a database if there isn't one yet
-CURRENT_TABLES="$(as_build "drush sql-query 'SHOW TABLES;'" /app/docroot)"
-if [ "$CURRENT_TABLES" == '' ] && [ -n "$DRUPAL_INSTALL_PROFILE" ]; then
-  as_build "echo 'y' | drush site-install '$DRUPAL_INSTALL_PROFILE'" "/app/docroot"
+set +e
+as_code_owner "drush sql-query 'SHOW TABLES;' | grep -v cache | grep -q ''" /app/docroot
+HAS_CURRENT_TABLES=$?
```
andytson-inviqa: is there any way to do this that wont run site-install if there are connectivity issues?

kierenevans: I'll remove the `set +e` so `drush sql-query` will fail if the database connection fails. This will then bubble up.

andytson-inviqa: grep returns non-zero if no match. I think you would have to deconstruct the command into separate commands

drupal8-apache/7.0/usr/local/share/drupal8/install_finalise.sh
```diff
@@ -45,11 +34,11 @@ set +e
...
 IS_NFS=$?
 set -e
-if [ "$IS_NFS" -ne 0 ]; then
-  chown -R www-data:www-data "$SETTINGS_DIR/files/"
+if [ "$IS_NFS" -ne 0 ] && [ -d "$SETTINGS_DIR/files/" ]; then
+  chown -R "$APP_USER:$APP_GROUP" "$SETTINGS_DIR/files/"
 fi
 
-as_build "drush cache-rebuild" "/app/docroot"
+as_code_owner "drush cache-rebuild" "/app/docroot"
```
Chapabu: This needs to come out. We're not installing the site here any more, so a cache rebuild causes the images to stop building.

sroze:
🚀

PR #52, opened 2016-11-28
Remove node sass from drupal.

kierenevans:


tonypiper:
do you still need nodejs installed in this?

kierenevans:
I think so, to avoid every project that requires node installing it themselves.

PR #53, opened 2016-11-28
Extract the running of confd to a separate script

kierenevans:
This is so we can re-use it in one-off scripts like drupal8's development install, which requires the correct database variables and user setup.

### DIFF
--- a/ubuntu/16.04/usr/local/share/bootstrap/run_confd.sh
+++ b/ubuntu/16.04/usr/local/share/bootstrap/run_confd.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Initialisation - Declare custom environment variables
+source /usr/local/share/env/custom_env_variables
+
+# Initialisation - Declare default environment variables
+source /usr/local/share/env/default_env_variables
+
+# Initialisation - Declare variables used by scripts in the ubuntu base image
+# to avoid undeclared variables!
+source /usr/local/share/env/bootstrap_env_variables
+
+# Initialisation - Pre templating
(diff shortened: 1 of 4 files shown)


ubuntu/16.04/usr/local/share/bootstrap/run_confd.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Initialisation - Declare variables used by scripts in the ubuntu base image
+# to avoid undeclared variables!
+source /usr/local/share/env/bootstrap_env_variables
+
+# Initialisation - Pre templating
+source /usr/local/share/bootstrap/pre_templating.sh
+
+# Initialisation - create a user to match the mountpoint's settings if told to
+source /usr/local/share/bootstrap/trigger_update_permissions.sh
```
andytson-inviqa: I think it might make sense to separate out the above sources from this file and include in the relevant areas. That way say for some reason a project wants to replace confd, they can replace this file without the above.

This was part of my reasons for going with bash functions in my build. It allows for more dynamic replacement, where in a scenario like this change it'd be

source /usr/local/share/bootstrap/setup.sh

do_confd

ubuntu/16.04/usr/local/share/bootstrap/run_confd.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+source /usr/local/share/bootstrap/setup.sh
```
kierenevans: Got you. I've updated the PR!

PR #55, opened 2016-11-29
Add basic auth capability

kierenevans:
Add basic authentication capability, so we can easily protect sites.


php-apache/7.0/README.md
```diff
@@ -3,3 +3,25 @@
...
+4. Provide this htpasswd line securely in the environment for this image as `BASIC_AUTH_HTPASSWD`
+5. Also provide the following variable with some values either through docker-compose environment or in
+   `/usr/local/share/env/custom_env_variables`:
+  ```
+  export BASIC_AUTH_ENABLED=true
+  ```
+6. You may also optionally configure the following in the same way:
+  ```
+  export BASIC_AUTH_REALM=Protected System
+  export BASIC_AUTH_TYPE=Basic
```
andytson-inviqa: BASIC_AUTH_TYPE basic seems implied, so perhaps this doesn't need to be configurable?

php-apache/7.0/usr/local/share/env/default_env_variables
```diff
@@ -12,3 +12,9 @@ export TIDEWAYS_API_KEY=${TIDEWAYS_API_KEY:-}
 export TIDEWAYS_CONNECTION=${TIDEWAYS_CONNECTION:-tcp://tideways:9135}
 
 export WORK_DIRECTORY=${WORK_DIRECTORY:-/app}
+
+export BASIC_AUTH_ENABLED=${BASIC_AUTH_ENABLED:-}
+export BASIC_AUTH_REALM=${BASIC_AUTH_REALM:-Protected System}
+export BASIC_AUTH_FILE=${BASIC_AUTH_FILE:-/etc/apache2/htpasswd}
+export BASIC_AUTH_TYPE=${BASIC_AUTH_TYPE:-Basic}
```
andytson-inviqa: That'll leave a lot of changes to names of files and inconsistencies with the nginx images. From a OS perspective it'd prob make sense tho, but would currently miss out `AuthDigestAlgorithm`

A prefix like "AUTH_HTTP_" might make more sense (I'd not recommend "HTTP_AUTH" though due to some languages treating HTTP_ as a request header lookup)

tonypiper: premature optimisation?

kierenevans: It's an easy change for the naming of variables - and better now before anything implements it. Pushing in a minute!

PR #56, opened 2016-12-02
Magento2 image permissions

kierenevans:
Very similar to the drupal8 permissions changes, allow for /app being a mountpoint with files that can't be chmodded or chowned. This is via using the "change user of web server" functionality when APP_USER_LOCAL is set to true.


magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento_finalise.sh
```diff
@@ -9,14 +9,25 @@ else
 fi ;
 
 source /usr/local/share/bootstrap/common_functions.sh
-bin/magento setup:upgrade
+
+cd /app || exit 1
+
+# Preserve compiled theme files across setup:upgrade calls.
+mkdir /tmp/assets
+cp -pR /app/pub/static/frontend/ /tmp/assets
```
sroze: If you manually precise the `cd /app` before, you can remove the `/app` in here then.

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento_finalise.sh
```diff
@@ -9,14 +9,25 @@ else
...
+chown -R "${CODE_OWNER}":"${CODE_GROUP}" pub/media pub/static var
+
+as_code_owner "bin/magento setup:upgrade"
+
+mv /tmp/assets/* /app/pub/static/frontend/
```
sroze: Same in here, you can remove the `/app`.

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento_finalise.sh
sroze: That way it'll be easier to potentially parameterize this.

sroze:
Sounds good to me 👍

PR #60, opened 2016-12-07
Additional safety for magento2 install

kierenevans:
Some assumptions were made regarding paths when setting this up, which do not hold true for more than one project. So, let's parameterise the paths affected by those assumptions.

This also adds extension points for the installation process, as per the drupal8 images.


sroze:
Looks VERY good to me 👍

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento.sh
```diff
@@ -28,9 +28,16 @@ fi
 
 if [ ! -d "vendor" ] || [ ! -f "vendor/autoload.php" ]; then
   as_code_owner "composer config repositories.magento composer https://repo.magento.com/"
-  as_code_owner "composer global config http-basic.repo.magento.com '$MAGENTO_USERNAME' '$MAGENTO_PASSWORD'"
-  as_code_owner "composer global config http-basic.toran.inviqa.com '$TORAN_USERNAME' '$TORAN_PASSWORD'"
-  as_code_owner "composer global config github-oauth.github.com '$GITHUB_TOKEN'"
+
+  if [ -n "$MAGENTO_USERNAME" ]; then
```
tonypiper: do we need to check that magento username and password are set? or could password be blank?

kierenevans: Good point :)

tonypiper: same for toran etc.

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento.sh
```diff
@@ -22,9 +28,16 @@ fi
...
-  as_code_owner "composer global config github-oauth.github.com '$GITHUB_TOKEN'"
+
+  if [ -n "$MAGENTO_USERNAME" ] && [ -n "$MAGENTO_PASSWORD" ]; then
+    as_code_owner "composer global config http-basic.repo.magento.com '$MAGENTO_USERNAME' '$MAGENTO_PASSWORD'"
+  fi
+  if [ -n "$GITHUB_TOKEN" ]; then
+    as_code_owner "composer global config github-oauth.github.com '$GITHUB_TOKEN'"
+  fi
+  if [ -n "$COMPOSER_CUSTOM_CONFIG_COMMAND" ]; then
+    as_code_owner "$COMPOSER_CUSTOM_CONFIG_COMMAND"
```
andytson-inviqa: I'd like to revisit this in a more customisable way, however for now it will do

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento.sh
```diff
@@ -35,30 +48,36 @@ if [ ! -d "vendor" ] || [ ! -f "vendor/autoload.php" ]; then
...
 fi
 
+mkdir -p pub/media pub/static var
 if [ "$IS_NFS" -ne 0 ]; then
   chown -R "${APP_USER}:${CODE_GROUP}" pub/media pub/static var
-  chmod ug+rw,o-w pub/media pub/static var
+  chmod -R ug+rw,o-w pub/media pub/static var
 else
-  chmod a+rw pub/media pub/static var
+  chmod -R a+rw pub/media pub/static var
```
andytson-inviqa: changing to facl in the future would be recommended here

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento_finalise.sh
```diff
@@ -2,29 +2,49 @@
...
+else
+  chmod a+rw pub/media pub/static var
+fi
+
 # Preserve compiled theme files across setup:upgrade calls.
-mkdir /tmp/assets
-cp -pR pub/static/frontend/ /tmp/assets
-chown -R "${CODE_OWNER}":"${CODE_GROUP}" pub/media pub/static var
+if [ -d pub/static/frontend/ ]; then
+  mkdir /tmp/assets
```
andytson-inviqa: I wonder if something closer to the original path would be better, in the case of if a volume is used, as moving between filesystems will be slow

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento.sh
kierenevans: Ideally it would be in the run_composer method that's in /usr/local/share/php/common_functions.sh - it's already a customisation to not run `--optimize-autoloader`. So yeah, would love to get this done properly!

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento.sh
kierenevans: setfacl doesn't work for docker volumes (on OSX, at least). So we can do it on CP where we aren't mounting code into the container but local is a problem.

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento_finalise.sh
kierenevans: Good point!

magento2-nginx/fpm-7.0/usr/local/share/magento2/install_magento.sh
andytson-inviqa: are you talking about volumes in the IS_NFS scenario? I'm just talking about when IS_NFS = 0

andytson-inviqa: ah right, I had thought the thing that was checking for the mount points would have made vboxsf show as IS_NFS=1

kierenevans: Ah yes, bad variable naming! `Users on /app type vboxsf (rw,nodev,relatime)`, ```root@fab42512d235:/app# is_nfs /app
root@fab42512d235:/app# echo $?
1```

PR #61, opened 2016-12-08
Update nginx base to use split configuration

andytson-inviqa:


andytson-inviqa:
Changes have been added to further customise the config and a readme added, ready for review

php-nginx/7.0/etc/confd/conf.d/nginx_site_phpfpm.conf.toml
```diff
@@ -0,0 +1,6 @@
+[template]
+src   = "nginx/site_base.conf.tmpl"
+dest  = "/etc/nginx/sites-available/default-30-rewriteapp.conf"
```
kierenevans: Should this be `default-30-phpfpm.conf` ?

andytson-inviqa: ok, fixed and giving expected output. Not much I can do with external testing for some of the env vars without the PR to generate self signed cert

PR #62, opened 2016-12-08
Update base images to use layered custom start scripts

andytson-inviqa:
This is so each layer can add their own additions without replacing the layer below, including the project


ubuntu/16.04/usr/local/bin/supervisor_start
```diff
@@ -8,6 +8,10 @@ source /usr/local/share/bootstrap/setup.sh
 source /usr/local/share/bootstrap/run_confd.sh
 
 # Initialisation - Runtime installation tasks
+for file in /usr/local/bin/supervisor_custom_start-*
```
andytson-inviqa: yes, it'll use the filesystem's sort order, the reason for the 20's

10 - web server
20 - framework

ubuntu/16.04/usr/local/bin/supervisor_start
```diff
@@ -8,6 +8,10 @@ source /usr/local/share/bootstrap/setup.sh
 source /usr/local/share/bootstrap/run_confd.sh
 
 # Initialisation - Runtime installation tasks
+for file in /usr/local/bin/supervisor_custom_start-*; do
+  source ${file}
```
andytson-inviqa: I seem to have to `# shellcheck source=/dev/null` to stop this being an issue, as there's no example one it can check

PR #63, opened 2016-12-08
Nginx self-signed certificate creation

andytson-inviqa:
depends on #61 and #62


andytson-inviqa:
I'd thought of adding optional env var cert/key data injection as well, but think this is enough for an iterative improvement

php-nginx/README.md
```diff
@@ -19,6 +19,15 @@ docker-compose build php56_nginx
 docker-compose push php56_nginx
 ```
 
+## About
+
+This is a Docker image for PHP (using the php-fpm SAPI) and Nginx. It uses, by
+default, our recommended default configuration, including:
+
+* A default HTTPS only website, with HTTP redirecting to HTTPS
```
sroze: AFAIK this PR do not introduce a HTTP to HTTPS redirection (which shouldn't IMHO), right?

andytson-inviqa: It should for new projects that want to gain from security and performance improvements. The default state is that HTTP redirects if HTTPS is enabled, but HTTP can be enabled as well

andytson-inviqa: This should be considered both for environments that don't have an SSL LB terminator, and those where the security requirements ask that inter-container traffic is TLS encrypted. Getting Vault to manage the TLS certs will help make that an easy reality for any project that choses to once the groundwork is done.

sroze: Agree with you, we should add a bold recommendation but not force that by default, as it wouldn't help newcomers... :/

sroze: > The default state is that HTTP redirects if HTTPS is enabled, but HTTP can be enabled as well

Where is that in your PR? 🤔

andytson-inviqa: It was introduced in a previous PR, that PR supported HTTPS as well, just was disabled by default until at least self signed certs were available.

#61 https://github.com/continuouspipe/dockerfiles/pull/61/files#diff-079ac36fbe458f15bd03fc001098e215R44

andytson-inviqa: https://github.com/continuouspipe/dockerfiles/pull/61/files#diff-17c9a37da5a06bfcb5ccfe32f06a3258R5 for the actual place this happens

sroze: Alright, so it was already merged so I believe that @kierenevans & you were okay with that. I'm not convinced at all by the UX but we'll see :)

andytson-inviqa: Can you elaborate that somewhere when you have the tim?

PR #65, opened 2016-12-12
Add the Symfony with nginx configuration

sroze:


php-apache/5.6/Dockerfile
```diff
@@ -0,0 +1,56 @@
+FROM quay.io/continuouspipe/ubuntu:16.04_v1
```
kierenevans: Double check this is from the latest master, as it would be `quay.io/continuouspipe/ubuntu16.04:v1.0`

symfony/3-nginx/etc/confd/templates/nginx/nginx_site.conf.tmpl
```diff
@@ -0,0 +1,27 @@
...
+
+    root /app/web;
+
+    server_name localhost;
+
+    location / {
+        try_files $uri /app.php$is_args$args;
+    }
+
+    location ~ ^/(app|app_dev|config)\.php(/|$) {
```
andytson-inviqa: once #60 is merged, only this location will be needed to be replaced. I'd be interested in considering how also minimising the replacement of this section in a good way

andytson-inviqa: obv changing the location to an env var would work, but might there be another way that's not too bad

andytson-inviqa:
Do you think there's a way we can do multiple PHP versions without duplicating configuration? Perhaps build ARGS? Specify the PHP version in an ARG that docker-compose can set?

andytson-inviqa:
#68 is introducing PHP versions as build ARGs, so you can use that to remove the duplication in this PR

symfony/3-nginx/etc/confd/templates/nginx/nginx_site.conf.tmpl
andytson-inviqa: I'd be ok with just the image replacing https://github.com/continuouspipe/dockerfiles/blob/master/php-nginx/etc/confd/templates/nginx/site_phpfpm.conf.tmpl with the updated location block short term if a decision can't be made. At least the duplication would only be contained to the base image and not projects that use it

sroze:
Already done by you guys, fab!

PR #67, opened 2016-12-13
Remove settings file completely from drupal8.

kierenevans:


sroze:
So does this configuration file needs to be created in all projects?

tonypiper:
@Chapabu any thoughts on this?

Chapabu:
There will need to be a config file of some sort on the project, but this is fairly normal for Drupal hosting (i.e Acquia needs specific settings, Platform.sh needs specific settings, etc).

So long as it's documented, it should be fine.

I'll be backporting settings from a working project into https://github.com/continuouspipe/drupal8-lightning at some point this week.

sroze:
Thanks @Chapabu. I've created the issue continuouspipe/drupal8-lightning#24 for you :)

PR #68, opened 2016-12-16
Add PHP 5.6 Image

kierenevans:
Using the PHP 7.0 image, but with build arguments to determine which version of PHP to install.

If 5.6 is chosen, Ondřej Surý's PPA will be used to provide the 5.6 packages, as ubuntu 16.04 ships with PHP 7.0 (🎉).

As we are using confd, we have to do a bit of manipulation of the .toml files in the build phase of the image, so that the right places will be written to when bringing up a container.


docker-compose.yml
```diff
@@ -74,17 +74,39 @@ services:
   php_apache:
```
andytson-inviqa: I think I want to suggest this updating to php70_apache (same with php_nginx)

kierenevans: I agree, done!

PR #71, opened 2016-12-17
Add memcached image.

kierenevans:
For Magento 1 sessions, and more!

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,6 +57,13 @@ services:
     depends_on:
       - varnish
 
+  memcached:
+    build:
+      context: ./memcached/1.4/
+    image: quay.io/continuouspipe/memcached1.4:v1.0
+    depends_on:
+      - ubuntu
+
   mysql:
     build:
       context: ./mysql/5.6/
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -57,6 +57,13 @@ services:
     depends_on:
       - varnish
 
+  memcached:
+    build:
+      context: ./memcached/1.4/
+    image: quay.io/continuouspipe/memcached1.4:v1.0
```
sroze: Can we add a `-` between `memcached` and the version number?

tonypiper: @sroze this looks consistent with the others...

sroze: yeah it looks like... don't like this name tho... @andytson-inviqa what's your PoV?

tonypiper: he's on leave until next year.

sroze: @andytson-inviqa, are you back?

andytson-inviqa: `quay.io/continuouspipe/memcached-1.4:v1.0` would prob be more preferable for me, but I don't think we should change it for just one image

andytson-inviqa: I think I prefer more quay.io/continuouspipe/memcached:1.4-1.0 (which is more like how rpms and debs are done)

sroze: In both cases that means that we need to rethink the entire image versioning to be consistent then... I agree with this PR and we can proceed with every image name in another PR then :)

PR #75, opened 2016-12-19
add Scala/sbt base image

svalaskevicius:


scala-base/1.0/Dockerfile
```diff
@@ -0,0 +1,51 @@
+FROM quay.io/continuouspipe/ubuntu:16.04_v1
```
sroze: @kierenevans & @andytson-inviqa named it to `quay.io/continuouspipe/ubuntu16.04:v1.0`

scala-base/1.0/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+  && apt-get install -y sbt \
+  && sbt sbtVersion \
+ \
+ # Clean the image \
+ && apt-get auto-remove -qq -y \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
+
+# Define working directory
+WORKDIR /root
```
sroze: Can we use `/app` as work directory as for all the other images?

scala-base/1.0/Dockerfile
svalaskevicius: is that tag working now? @kierenevans suggested to use this one

scala-base/1.0/Dockerfile
svalaskevicius: ofc, although is app meant to be for the deployed application or it's sources? its meant to be compiled in the image, deployed as binaries and then sources can be removed.. so the deploy dir should be different from the sources / compilation base

svalaskevicius:
wow. does the linter need to be kicked? :D

`Expressions don't expand in single quotes, use double quotes for that` - I meant them not to be expanded.

scala-base/1.0/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+MAINTAINER Sarunas Valaskevicius <svalaskevicius+cp-dockerfiles@inviqa.com>
+
+# based on java:8 and https://github.com/hseeberger/scala-sbt images
+
+ENV SBT_VERSION 0.13.13
+
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
+ENV JAVA_VERSION 8u111
+ENV JAVA_DEBIAN_VERSION 8u111-b14-2ubuntu0.16.04.2
```
andytson: why hard-code the version in? we want security updates to happen as part of release cycle

svalaskevicius: taken from the dockerfile I looked at.. https://github.com/docker-library/openjdk/blob/e6e9cf8b21516ba764189916d35be57486203c95/8-jdk/Dockerfile don't mind changing that though as had a similar question

svalaskevicius: ah they have this https://github.com/docker-library/openjdk/blob/master/update.sh ... will update to use just java8..

svalaskevicius:
@sroze what do you think now?

svalaskevicius:
cc @andytson

PR #78, opened 2017-01-04
Fix socket path to be /run/php7.0-fpm.sock rather than /run/php/php7.0-fpm.sock

kierenevans:
If kept in /run/php, an extra mkdir for /run/php is required before launching.


sroze:
Don't we have to change the PHP-FPM configuration at the same time then?

kierenevans:
The php-fpm configuration is set to launch on /run/php7.0-fpm.sock already, so this was pointing to the wrong place.

PR #79, opened 2017-01-04
Fix PHP installation script location.

kierenevans:
As called by:
```
 ➜  dockerfiles git:(feature/fix-php-installation-script-location) ✗ grep -R php/install.sh .
./php-nginx/usr/local/bin/supervisor_custom_start-10-php-nginx:bash /usr/local/share/php/install.sh```


sroze:
If you say so 👍

PR #80, opened 2017-01-05
Remove http2 from port 80

kierenevans:
Accessing http was causing a binary download due to a http2 handshake trying to occur.

### DIFF
--- a/php-nginx/etc/confd/templates/nginx/site.conf.tmpl
+++ b/php-nginx/etc/confd/templates/nginx/site.conf.tmpl
@@ -1,6 +1,6 @@
 server {
     server_name _;
-    listen {{ getenv "WEB_HTTP_PORT" }} default_server http2;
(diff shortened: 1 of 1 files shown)


php-nginx/etc/confd/templates/nginx/site.conf.tmpl
```diff
@@ -12,10 +12,13 @@ server {
     server_name _;
 {{ end }}
 {{ if eq "true" (getenv "WEB_HTTPS_OFFLOADED") }}
-    listen {{ getenv "WEB_HTTPS_PORT" }} default_server http2;{{ else }}
+    listen {{ getenv "WEB_HTTPS_PORT" }} default_server http2;
+{{ else }}
```
andytson-inviqa: the {{ else }} was on the same line to avoid multiple newlines

PR #83, opened 2017-01-06
Convert usr/local/share/env to use alphanumeric wildcard importing

andytson-inviqa:
This allows any number of env files to be added and imported, avoiding assumptions in image heirarchy setup


andytson-inviqa:
will be best testing this

andytson-inviqa:
I missed committing the test.sh change, updated now

PR #89, opened 2017-01-12
Create parent directory for the parameters.yml before attempting to write to it.

kierenevans:
Fixes:
```
/usr/local/share/symfony/symfony_functions.sh: line 11: /app/app/config/parameters.yml: No such file or director
```


andytson-inviqa:
Extremely unorthodox for a Symfony app to not have this directory, but say so, it doesn't hurt to ensure it exists

andytson-inviqa:
chances are the parameters.yml wont even be read if there's no config directory

PR #90, opened 2017-01-13
Chown to build user before composer runs

kierenevans:
Should fix the inability for composer running as the build user to create /app/bin, for example.

### DIFF
--- a/php-apache/usr/local/share/php/common_functions.sh
+++ b/php-apache/usr/local/share/php/common_functions.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+do_build_permissions() {
+  if [ "$IS_NFS" -ne 0 ]; then
+    setfacl -R -m "d:u:$CODE_OWNER:rwX" -m "u:$CODE_OWNER:rwX" /app/
+  else
+    chown -R "$CODE_OWNER:$CODE_GROUP" /app/
+  fi
+}
+
 run_composer() {
   if [ ! -d "/app/vendor" ] || [ ! -f "/app/vendor/autoload.php" ]; then
     mkdir -p /app/vendor
--- a/symfony-nginx/usr/local/share/symfony/symfony_functions.sh
+++ b/symfony-nginx/usr/local/share/symfony/symfony_functions.sh
@@ -10,7 +10,6 @@ do_symfony_build_permissions() {
     # Prepare a default parameters.yml. incenteev/parameters-handler can still update it
     mkdir -p /app/app/config
     [ ! -f /app/app/config/parameters.yml ] && echo 'parameters: {}' > /app/app/config/parameters.yml
-    setfacl -R -m "d:u:$CODE_OWNER:rwX" -m "u:$CODE_OWNER:rwX" /app/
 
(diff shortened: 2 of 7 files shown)


php-apache/usr/local/share/php/common_functions.sh
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+do_build_permissions() {
+  if [ "$IS_NFS" -ne 0 ]; then
+    setfacl -R -m "d:u:$CODE_OWNER:rwX" -m "u:$CODE_OWNER:rwX" /app/
```
andytson-inviqa: sadly GCE doesn't support setfacl, as we just found out, so this will have to be chown

php-apache/usr/local/share/php/common_functions.sh
```diff
@@ -1,5 +1,13 @@
...
+  else
+    chown -R "$CODE_OWNER:$CODE_GROUP" /app/
```
andytson-inviqa: I think this wont work on some volume types, where the symfony image is currently doing `chmod +rw` instead

kierenevans:
Ready for re-review

symfony-nginx/usr/local/share/symfony/symfony_functions.sh
```diff
@@ -1,23 +1,26 @@
...
-
     # Fix permissions so the web server user can write to cache and logs folders
     if [ "$SYMFONY_MAJOR_VERSION" -eq 2 ]; then
-      setfacl -R -m "d:u:$APP_USER:rwX" -m "u:$APP_USER:rwX" /app/app/{cache,logs}
-      chmod -R ug+rw,o-rw /app/app/{cache,logs}
+      chown -R "$APP_USER:$CODE_GROUP" /app/app/{cache,logs}
+      chmod -R ug+rw,o-rwx /app/app/{cache,logs}
     fi
-    setfacl -R -m "d:u:$APP_USER:rwX" -m "u:$APP_USER:rwX" /app/var
+    chown -R "$APP_USER:$CODE_GROUP" /app/var
```
andytson-inviqa: we'll need to move this to after composer, due to composer creating files that are build:build rw-r-r

andytson-inviqa: also with it being moved, chmoding g+w is not needed

kierenevans: It got moved to after composer via the ordering in symfony-nginx/usr/local/share/container/baseimage-30.sh

kierenevans: g+w allows you to run console commands as the build user, though this may not be required

andytson-inviqa: oh I see, but composer will need the chown build:build /app, which is why I was suggesting two functions

andytson-inviqa: nvm, I tldr'd the PR

PR #94, opened 2017-01-16
Move literal env var defaults to Dockerfiles

andytson-inviqa:


andytson-inviqa:
@sroze @kierenevans what do you think of this idea?

sroze:
The problem I can see is that we would see the default values noise in the image layers history, and it would hint people that they could do the same for their child images, thing that they should put the environments at build time, thing that they should not.

andytson-inviqa:
Yep true, however as these ones are inspectable, it's a way of self-documenting standard what env vars are available.

I don't really mind not using it for this, but will need to use it for SYMFONY_ENV, as it needs to be persisted to the build image, and also used by default on run. This is the easiest way to achieve that

andytson-inviqa:
if it was cut up into build-time env vars vs run-time env vars, would that be more acceptable?

sroze:
Definitely, if we can separate the different types of variables, that's fine to me then! 👍

andytson-inviqa:
actually, having tried to do that, I've decided against it

PR #95, opened 2017-01-16
Update Magento 2 image to use function approach

kierenevans:
Updating the Magento2 image to use the new function based approach by @andytson-inviqa 

WIP, probably won't build yet.


magento2-nginx/fpm-7.0/usr/local/share/container/baseimage-30.sh
```diff
@@ -12,17 +14,25 @@ do_development_start() {
   do_magento2_development_start
 }
 
+alias_function do_templating do_magento2_templating_inner
+do_templating() {
+  mkdir -p /home/build/.hem/gems/
+  chown -R build:build /home/build/.hem/
```
andytson-inviqa: you can fix this issue by doing `useradd build --create-home`, then Hem will be able to create the directory

kierenevans: https://github.com/continuouspipe/dockerfiles/blob/master/ubuntu/16.04/Dockerfile#L35 sets this up but it didn't seem to help, unless this is a leftover from before that was added. I'll try it without when I get around to rebuilding.

kierenevans:
A `quay.io/continuouspipe/magento2-nginx-php7:dev-functionised` image tag has been pushed up for someone to test. Might be me when I have some time :)

magento2-nginx/fpm-7.0/usr/local/share/magento2/magento_functions.sh
```diff
@@ -0,0 +1,245 @@
...
+    done
+  fi
+}
+
+function do_magento_cache_flush() {
+  # Flush magento cache
+  as_code_owner "bin/magento cache:flush"
+}
+
+function do_magento_create_web_writable_directories() {
```
kierenevans: This double version of do_magento_create_web_writable_directories (see line 18) stops bin/magento from working.

kierenevans:
The PR that this is based on, #181, currently has a bug with truthy values and confd. Placing this on hold until resolved.

PR #98, opened 2017-01-16
Add postfix as a sendmail (cli interface) SMTP relay

andytson-inviqa:
PHP's mail() function uses sendmail (cli interface) to send mail. We don't want the container to send mail directly, however some projects send mail during the web request without failure handling.

If a project sends mail via a queue or cron and is handling failures, it's acceptable for it to talk to the 3rd-party mail provider service directly.

This adds postfix only for sendmail (php mail();) relaying, and is disabled by default

Setting SENDMAIL_RELAY_HOST activates the postfix supervisor service, and other environment variables control it's authentication and security level


php-apache/etc/confd/templates/postfix/main.cf.tmpl
```diff
@@ -0,0 +1,24 @@
+append_dot_mydomain = no
+biff = no
+header_size_limit = 4096000
+inet_interfaces = loopback-only
```
kierenevans: IPv6 DNS resolution on Amazon was a problem for us, so we had to use IPv4 only. Fix from @fratcliffe incoming, though fixing IPv6 DNS might be the best option.

php-nginx/usr/local/sbin/postfix.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# call "postfix stop" when exiting
+trap "{ echo Stopping Postfix; /usr/sbin/postfix stop; exit 0; }" EXIT
+
+source /etc/sysconfig/network
+
+/usr/libexec/postfix/aliasesdb
+/usr/libexec/postfix/chroot-update
+
+if [ -n "${POSTFIX_RELAY_USER}" ]; then
```
kierenevans: Fixes incoming from @fratcliffe but this would be `SENDMAIL_RELAY_USER`, we think.

php-apache/usr/local/sbin/postfix.sh
```diff
@@ -0,0 +1,18 @@
...
+
+source /etc/sysconfig/network
+
+/usr/libexec/postfix/aliasesdb
+/usr/libexec/postfix/chroot-update
+
+if [ -n "${POSTFIX_RELAY_USER}" ]; then
+  postmap /etc/postfix/sasl_passwd
+fi
+
```
kierenevans: DNS resolution was a problem for us, so we had to copy /etc/resolv.conf to somewhere visible by postfix. Fix from @fratcliffe incoming

php-apache/etc/confd/templates/postfix/main.cf.tmpl
```diff
@@ -0,0 +1,24 @@
...
+myorgin = $myhostname
+{{ if getenv "SENDMAIL_RELAY_HOST" }}
+relayhost = [{{ getenv "SENDMAIL_RELAY_HOST" }}]:{{ getenv "SENDMAIL_RELAY_PORT" }}
+{{ end }}
+{{ if getenv "SENDMAIL_RELAY_USER" }}
+smtp_sasl_auth_enable = yes
+smtp_sasl_password_maps = hash:/etc/postfix/sasl_passwd
+smtp_sasl_security_options = noanonymous
+{{ end }}
+smtp_tls_CAfile = /etc/pki/tls/cert.pem
```
kierenevans: This path doesn't exist on ubuntu, fix from @fratcliffe incoming

php-apache/etc/confd/templates/postfix/main.cf.tmpl
```diff
@@ -0,0 +1,24 @@
...
+smtp_tls_security_level = {{ getenv "SENDMAIL_RELAY_TLS_SECURITY_LEVEL" }}
+smtp_tls_session_cache_database = btree:${data_directory}/smtp_scache
+smtpd_tls_CAfile = /etc/pki/tls/cert.pem
```
kierenevans: This path doesn't exist on ubuntu, fix from @fratcliffe incoming

php-apache/usr/local/sbin/postfix.sh
```diff
@@ -0,0 +1,18 @@
...
+# start postfix
+/usr/sbin/postfix - c /etc/postfix start
```
kierenevans: Extra space between - and c, fix incoming from @fratcliffe

php-nginx/etc/confd/templates/postfix/main.cf.tmpl
```diff
@@ -0,0 +1,24 @@
+append_dot_mydomain = no
+biff = no
+header_size_limit = 4096000
+inet_interfaces = loopback-only
```
kierenevans: IPv6 DNS resolution on Amazon was a problem for us, so we had to use IPv4 only. Fix from @fratcliffe incoming, though fixing IPv6 DNS might be the best option.

php-apache/usr/local/sbin/postfix.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# call "postfix stop" when exiting
+trap "{ echo Stopping Postfix; /usr/sbin/postfix stop; exit 0; }" EXIT
+
+source /etc/sysconfig/network
+
+/usr/libexec/postfix/aliasesdb
+/usr/libexec/postfix/chroot-update
+
+if [ -n "${POSTFIX_RELAY_USER}" ]; then
```
andytson: I vote for POSTFIX so as to match the service name, as SENDMAIL is incorrect term and serves only to confuse people further in regards to differences between sendmail software and sendmail cli interface

andytson: ah I see, I've mixed and matched env variable namings here

kierenevans:
I've rebased this on top of master and imported the fixes we made when re-implementing this PR elsewhere. As this works elsewhere, it most likely works here!

PR #99, opened 2017-01-17
Merge symfony-3-apache into symfony-nginx work

andytson-inviqa:
Using template variables to control the base image and etc directory to use

Note this changes the registry push location to match the format of the symfony-nginx repositories.

It also brings in PHP 7.1 and 5.6 versions of Symfony for Apache


andytson-inviqa:
@sroze I did add them already in the first commit

PR #101, opened 2017-01-17
Ensure var directory writable by composer

### LINKED ISSUE #100: Add `var` in `.dockerignore` breaks the Symfony base image
When the `var` directory is in the `.dockerignore` file, the `build` user can't access the `/app/var` directory apparently during the build of the Docker image.

kierenevans:
Fixes #100


sroze:
Wow, that was fast 🎉 👍

PR #104, opened 2017-01-17
Add optional composer flags

kierenevans:
Adds `--no-dev` if not in DEVELOPMENT_MODE, though you can influence it at a higher image level by declaring COMPOSER_INSTALL_FLAGS.

This can be used by the symfony image for example, to disable the update parameters plugin.


andytson-inviqa:
I'm going to test this now as I need to disable the plugins on a project

andytson-inviqa:
works as expected

sroze:
That is not ideal for the remote development... I believe we should opt-out.

sroze:
That is not ideal for the remote development... I believe we should make this optionable using an environment variable.

kierenevans:
It is, DEVELOPMENT_MODE :)

sroze:
What I mean is that by default, it is not using `--no-dev`.

PR #105, opened 2017-01-18
Remove http2 from another nginx http port conf

andytson-inviqa:
Nginx doesn't support http/1.1 upgrade to 2.0 on standard http


sroze:
Yes, definitely 👍

PR #106, opened 2017-01-19
Move COMPOSER_INSTALL_FLAGS to after 50-bootstrap

andytson-inviqa:
This fixes the issue of DEVELOPMENT_MODE not being set automatically before this is run


php-nginx/usr/local/share/env/55-stack
```diff
@@ -0,0 +1,6 @@
+DEFAULT_COMPOSER_FLAGS="--no-interaction --optimize-autoloader"
```
kierenevans: Do source'd files need a hashbang?

andytson-inviqa: they don't for sourcing, but for format detection they are needed, so I've added them

PR #107, opened 2017-01-19
Add to php-apache HTTPS and apache template splitting.

andytson-inviqa:
Note:
* it turns on HTTPs by default and redirects HTTP, so technically not BC, but BC maintained by setting WEB_HTTPS=false or WEB_HTTP=true


andytson-inviqa:
actually, it's only drupal, and that one's simple, so I'll update that

andytson-inviqa:
ok, Drupal images now use the templates as well, ready for review

PR #109, opened 2017-01-20
Do not call finalise script from the build part, as it needs external services like a database.

kierenevans:
For magento 2, similar to the one for drupal 8 in #103


sroze:
@andytson-inviqa happy to merge?

PR #110, opened 2017-01-20
Add a cron supervisor service that starts if START_MODE=cron

andytson-inviqa:


andytson-inviqa:
I wish there was a decent cron we can trust that allowed logging everything to stdout/err

sroze:
Why not moving this cron directly to the ubuntu image?

andytson-inviqa:
@sroze I'm happy to move it there if others are

sroze:
@kierenevans what do you think?

kierenevans:
Images without web servers or php will probably still need the ability to run a scheduled task somehow.
I am wary of moving it up to our ubuntu image in terms of adding clutter, but on the other hand we shouldn't need to solve something more than once.

andytson-inviqa:
how about a intermediate layer?

andytson-inviqa:
I've moved it to the Ubuntu image now. It's not much clutter as it's just the supervisor config, and cron is already installed in the ubuntu Dockerfile

PR #111, opened 2017-01-21
Add Mailcatcher image build

andytson-inviqa:


sroze:
The linter said that you need to precise a maintainer :)

> Specify a maintainer of the Dockerfile

andytson-inviqa:
sorted, ready for review (I changed it also to run from supervisord and use standard SMTP/HTTP ports)

tonypiper:
Just a question - do we need to protect this from drive-by port-scanners?

kierenevans:
For CP, I would recommend not exposing the ports to the world. If on a trusted kubenetes cluster, it should be alright.
Outside CP, yes we would need some sort of protection.

andytson-inviqa:
@tonypiper I was thinking for another PR to have an Nginx without PHP layer that could be used to protect if actually exposed (well unless httplabs will be done and done as a container for the only exposed ports of a project)

PR #117, opened 2017-01-24
Update M2 image to use redis for session storage and confd for configs

hgajjar-sessiondigital:
- Added magento config files to confd
- Use Redis for session storage
- Fix nginx config for default mage run type


magento2-nginx/fpm-7.0/etc/confd/conf.d/magento_env.php.toml
```diff
@@ -0,0 +1,8 @@
+[template]
+src   = "magento/env.php.tmpl"
+dest  = "/app/app/etc/env.php"
```
kierenevans: Might need to `mkdir -p /app/app/etc/` before templating runs, just in case. See https://github.com/continuouspipe/dockerfiles/blob/master/drupal8-apache/7.0/usr/local/share/container/baseimage-30.sh#L15-L19

hgajjar-sessiondigital:
@sroze @kierenevans variables are documented, can you please review?

PR #119, opened 2017-01-24
Feature/elasticsearch

blongden:


elasticsearch/2.4/README.md
```diff
@@ -0,0 +1,15 @@
+# ELASTICSEARCH 2.4
+
+```Dockerfile
+FROM quay.io/continuouspipe/elasticsearch:2.4
```
kierenevans: `FROM quay.io/continuouspipe/elasticsearch2.4:stable`

docker-compose.yml
```diff
@@ -178,6 +178,13 @@ services:
     depends_on:
       - external_redis
 
+  elasticsearch:
+    build:
+      context: ./elasticsearch/2.4/
```
andytson-inviqa: I'd recommend dropping this 2.4 off. If we want different versions, we'll template it the same way the mongodb and mysql ones are done

andytson-inviqa: I mean, storing it all in ./elasticsearch/ rather than ./elasticsearch/2.4/

blongden: Oh, just using Dockerfile-2.4 etc?

andytson-inviqa: You can leave that as Dockerfile for now unless you want to add other versions

andytson-inviqa: the other Dockerfile-*'s are only committed as CP doesn't have a pre-build step that can run build.sh

blongden: Gotcha.

kierenevans:
In advance of the PR being merged, I have created https://quay.io/repository/continuouspipe/elasticsearch2.4 and the CP build robot has been added with write permissions.

PR #133, opened 2017-02-01
Fix Magento2 hem check

rocketage:


kierenevans:
There's one more in the install_finalise.sh script

PR #134, opened 2017-02-01
Add ability to use Symfony web/app_*.php files

andytson-inviqa:
… whilst restricting them by default


symfony/usr/local/share/env/45-framework
```diff
@@ -10,3 +10,4 @@ fi
 
 export SYMFONY_MAJOR_VERSION=${SYMFONY_MAJOR_VERSION:-${SYMFONY_MAJOR_VERSION_GUESS}}
 export SYMFONY_CONSOLE=${SYMFONY_CONSOLE:-${SYMFONY_CONSOLE_DEFAULT}}
+export SYMFONY_APP_ENDPOINT_REGEX=$(php -r 'echo preg_quote(getenv("SYMFONY_APP_ENDPOINT"));')
```
kierenevans: The $() will need to be in quotes

symfony/nginx/etc/confd/templates/nginx/site_disable_further_php.conf.toml
```diff
@@ -0,0 +1,3 @@
+location ~ ^/.*\.php(/|$) {
```
kierenevans: File will need to be .tmpl rather than .toml

symfony/usr/local/share/env/45-framework
andytson-inviqa: I've checked and ENV_VAR=$(echo test test) and ENV_VAR=${VAR:-test test} appear to still treat the value as one argument, so the quotes aren't needed

andytson-inviqa:
ok, rebased with the env file shellcheck issue fixed, and Travis is happy

PR #135, opened 2017-02-01
Fix issue with test.sh not testing env files

andytson-inviqa:


andytson-inviqa:
well, I'm surprised this passed first time

PR #140, opened 2017-02-03
Create a phantomjs2 image

andytson-inviqa:


andytson-inviqa:
I don't see any official image, just loads of unofficial ones

sroze:
Fair enough then :)

PR #144, opened 2017-02-06
Add cron to magento2

kierenevans:
Start with a dedicated entry in docker-compose.yml, passing "START_MODE=cron" as an environment variable


andytson-inviqa:
any chance this could be supervisorised so that console output gets sent to docker logs?

PR #149, opened 2017-02-08
Remove hem and nodejs from web images

kierenevans:
Remove hem and nodejs from the web images, to make use of ContinuousPipe's new artefact building system instead.

I imagine that users of https://github.com/continuouspipe/remote-environment-client for remote dev will want to keep nodejs in for frontend builds.
Perhaps we should keep nodejs in or find a way around pre-installing in the image that gets used for other non-remote-dev environments?

Until #95 lands, the magento2 image will keep hem in.


sroze:
I'd close this PR, I'm afraid it'll be merged and it'll break every child image...

PR #150, opened 2017-02-08
Clear magento2 cache before setup upgrade

kierenevans:
Needed as redis or the generated folder may a different version of magento cached. For example if upgrading versions. This causes "bin/magento" calls to fail to boot magento.


kierenevans:
Confirmed as working when upgrading magento!

sroze:
Based on the discussions on the G project, I'll take as granted that it's working :)

PR #152, opened 2017-02-09
Add capability to whitelist IPs to bypass authentication.

kierenevans:
Requires testing but allow IP addresses to bypass authentication.
Starts off with localhost to allow easier curl testing.


andytson-inviqa:
I've done testing of both apache and nginx versions when I made my changes. Works as expected to me after some fixes I did.

PR #157, opened 2017-02-10
Allow setting mage run type in magento2

kierenevans:
Includes #156 as I wanted to build :latest to test stuff out...


kierenevans:
Updated docs!

PR #163, opened 2017-02-14
Make app mount point detection more generic

dcole-inviqa:
any chance this can be made a bit more generic and not as coupled to vbox?

### DIFF
--- a/ubuntu/16.04/usr/local/share/bootstrap/common_functions.sh
+++ b/ubuntu/16.04/usr/local/share/bootstrap/common_functions.sh
@@ -59,18 +59,13 @@ is_hem_project() {
 }
 
 is_app_mountpoint() {
-  grep -q -E "/app (nfs|vboxsf|fuse\.osxfs)" /proc/mounts
+  findmnt $WORK_DIRECTORY
   return $?
 }
 
 is_chown_forbidden() {
   # Determine if the app directory is an NFS mountpoint, which doesn't allow chowning.
   grep -q -E "/app (nfs|vboxsf)" /proc/mounts
-  return $?
-}
-
-is_vboxsf_mountpoint() {
-  grep -q "/app vboxsf" /proc/mounts
   return $?
 }
 
--- a/ubuntu/16.04/usr/local/share/env/50-bootstrap
+++ b/ubuntu/16.04/usr/local/share/env/50-bootstrap
@@ -10,16 +10,13 @@ set +e
 is_chown_forbidden
 export IS_CHOWN_FORBIDDEN=$?
 
-is_vboxsf_mountpoint
-export IS_VBOXSF_MOUNTPOINT=$?
-
 is_app_mountpoint
 export IS_APP_MOUNTPOINT=$?
 set -e
 
 export DEVELOPMENT_MODE=${DEVELOPMENT_MODE:-${IS_APP_MOUNTPOINT}}
 
-if [ "$IS_VBOXSF_MOUNTPOINT" -eq 0 ]; then
+if [ "$IS_APP_MOUNTPOINT" -eq 0 ]; then
   export APP_USER_LOCAL=${APP_USER_LOCAL:-true}
 else
   export APP_USER_LOCAL=${APP_USER_LOCAL:-false}


ubuntu/16.04/usr/local/share/bootstrap/common_functions.sh
```diff
@@ -69,11 +69,6 @@ is_chown_forbidden() {
   return $?
 }
 
-is_vboxsf_mountpoint() {
-  grep -q "/app vboxsf" /proc/mounts
-  return $?
-}
-
```
andytson-inviqa: we don't want to treat fuse.osxfs the same way as vboxsf, so need this to be a separate check

kierenevans: The separate treatment is to work around not being able to chown or vboxsf or NFS, or even chmod on vboxsf!

andytson-inviqa: for instance osxfs has none of the chmod/chown issues as the other mount points, but still want to treat it as development mode by default. There are similar reasons for nfs, for the APP_USER_LOCAL isn't needed for it

dcole-inviqa: use case would be im using native linux (no vm at all) so the only change i need when in development is the uid/gid fix.

root@fd0b312508fa:/# findmnt /app

TARGET SOURCE                                 FSTYPE OPTIONS
/app   /dev/sda1[/home/david/projects/forces] ext4   rw,relatime,errors=remount-ro,data=ordered

by being specific to vbox/osxfs etc those env variables are never set. how about starting with a check like in the PR then running the edge cases to account for chown based on specific filesystem types.

am happy to adjust the PR if it would be accepted.

andytson-inviqa: so in summary, you want APP_USER_LOCAL enabled so the ext4 mounted owner/group is that of your host. I think it might be dangerous doing that by default as if volume mounts are sometimes used in non-development environments (not really us but others). APP_USER_LOCAL can be set in the docker-compose.yml as false, but your host bash profile does `export APP_USER_LOCAL=true` to enable it for yourself only.

dcole-inviqa: I dont think thats a valid solution, although volume mounts might be used in production i'd imagine/hope the root $WORK_DIRECTORY (/app) would only be mounted by developers and if mounting /app in production is needed then the onus should be on that person and not developers to change the environment variable. 

otherwise youre asking developers to add CP specific variables to their bash profiles which if refactored would break those people environments and force them to monitor base image changes...

if you believe people are likely to be mounting /app in production then i can see a few approaches.

1. favour the developer as the person creating production image is more likely to track and follow base image changes so they can set the appropriate variable if they really need to mount /app

2. check the mount for the home|user prefix if present its a developer

3. introduce a PRODUCTION variable, when this is set skip over the smart logic of trying to guess vbox/osxfs/fuse/linux/chown/chmod etc....

dcole-inviqa: PS. if one approach is favoured am happy to adjust this PR :)

andytson-inviqa: btw, there's nothing CP-specific about it, it's a docker solution

andytson-inviqa: we're just considering whether our original preferred generic fix (setfacl) would solve this without the need for an ext4 APP_USER_LOCAL

dcole-inviqa: By CP specific i meant these are the CP Dockerfiles so they are specific to this project :) ... ideally a dev should just do 'docker-compose up -d' and be done.

If theres an alternate approach with setfacl, ive never needed to use it so am not that familiar with. is there a timeframe? if its a maybe or at the back of the backlog would one of the above approaches be okay?

sroze:
I think that this as been changed already, closing :)

PR #166, opened 2017-02-14
Fix nginx access_log directive by specifying default format

andytson-inviqa:


andytson-inviqa:
I've tested this in the container that was failing, and it actually works finally

PR #167, opened 2017-02-15
Allow args to passthru to tasks

dcole-inviqa:
can we allow args to be preserved and passed through to the task being called.

### DIFF
--- a/ubuntu/16.04/usr/local/bin/container
+++ b/ubuntu/16.04/usr/local/bin/container
@@ -2,6 +2,7 @@
 set -xe
 
 TASK=$1
+ARGS=$@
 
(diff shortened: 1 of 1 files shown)


andytson-inviqa:
I was thinking of doing this also, but wasn't decided on whether to instead allow multiple do_ functions to be specified instead of function arguments

dcole-inviqa:
fixed the shellcheck issues if you decide to go with with this approach, feel free to close if you go with the other approach :)

dcole-inviqa:
as a compromise i guess you could have a core task which runs multiple....

do_chain() { 
 for (task in $@)
 do
   "do_$task"
 done
}

andytson-inviqa:
yep, sounds good

PR #168, opened 2017-02-15
Add a nginx-reverse-proxy image.

andytson-inviqa:
This adds an ability to define a reverse proxy from an environment variable


andytson-inviqa:
Quay.io repo created and bot user assigned

PR #169, opened 2017-02-16
Feature/migrate drupal8 image to functions

Chapabu:
* Removes unnecessary tasks from the Drupal 8 image 
* Migrates to functions instead of scripts


andytson-inviqa:
thanks for this. We'd like to try to maintain some backwards-compatibility with the bits that projects are likely to edit, such as the _custom.sh scripts

drupal8-apache/7.0/usr/local/share/container/baseimage-30.sh
```diff
@@ -1,41 +1,51 @@
...
+#####
+alias_function do_build do_drupal_build_inner
+do_build() {
+  do_drupal_build_inner
+  do_drupal_build
 }
 
-do_drupal8_install() {
-  bash /usr/local/share/drupal8/install.sh
+alias_function do_build do_drupal_development_build_inner
```
kierenevans: This is adding to the one declared above on line 17, which might not be what you want. If it isn't then `do_drupal_development_build` in the drupal_functions.sh and the `do_development_build` below can be removed.

drupal8-apache/7.0/usr/local/share/drupal/drupal8_functions.sh
```diff
@@ -0,0 +1,113 @@
...
+#####
+# Perform tasks on container start
+#####
+do_drupal_start() {
+  do_drupal_permissions
+  do_drupal_legacy_install_finalise_script
+}
+
+do_drupal_development_start() {
+  do_drupal_start
```
kierenevans: `check_development_start` that triggers `do_development_start` when in dev mode is triggered from `do_start` originally (see https://github.com/continuouspipe/dockerfiles/blob/0e9a3f71562952f50774772c4faa2c53be20bb1e/ubuntu/16.04/usr/local/share/container/baseimage-10.sh#L15 )

That means the do_drupal_start will get triggered anyway, so no need to repeat.

If `do_drupal_build` where here instead, we would re-run the dev install steps on container start which is what we want for local development when /app is a mountpoint.

drupal8-apache/7.0/usr/local/share/drupal/drupal8_functions.sh
```diff
@@ -0,0 +1,113 @@
...
+  mkdir -p "${WEB_DIRECTORY}/sites/default/files"
+}
+
+#####
+# Fix directory permissions.
+#####
+do_drupal_permissions() {
+  if [ "$IS_CHOWN_FORBIDDEN" -ne 0 ]; then
+    # Give the docroot and config directories to the web user.
+    chown -R "${APP_USER}":"${APP_GROUP}" "${WEB_DIRECTORY}"
```
kierenevans: Please update to be specific directories if /app/docroot/* is not needed. Any directory owned by the web server can be modified in the event of a security issue being exploited. We want to reduce the surface area of attack as much as possible.

drupal8-apache/7.0/usr/local/share/drupal/drupal8_functions.sh
```diff
@@ -0,0 +1,113 @@
...
+do_drupal_development_install() {
+
+  if [ "$INSTALL_DRUPAL" == 'false' ]; then
+    return
+  fi
+
+  # Drop the database if we need to force the install every time.
+  if [ "$FORCE_DATABASE_DROP" == 'true' ]; then
+    echo 'Dropping the Drupal DB if it exists'
+    "drush sql-drop -y -r ${WEB_DIRECTORY}"
```
kierenevans: Should drush be running as root?

kierenevans: Oh, this isn't being executed, which is why it wasn't installing!

If drush should run as root (quotes only around the variable):
`drush sql-drop -y -r "${WEB_DIRECTORY}"` 

If drush should not run as root (ideal solution):
`as_code_owner "drush sql-drop -y -r ${WEB_DIRECTORY}"` 

If the build user does not have permissions, you can either choose to reset permissions before/after or run with `as_app_user` to run as the web server user, though beware any files created will be owned by the web server user.

drupal8-apache/7.0/usr/local/share/drupal/drupal8_functions.sh
```diff
@@ -0,0 +1,113 @@
...
+  fi
+
+  # If we're supposed to install Drupal, and it's not currently installed - then install it.
+  if ! drush status bootstrap | grep -q Successful ; then
+    echo 'Installing Drupal'
+    "drush site-install ${DRUPAL_INSTALL_PROFILE} -y -r ${WEB_DIRECTORY}"
```
kierenevans: Ditto here for the drush command

drupal8-apache/7.0/usr/local/share/drupal/drupal8_functions.sh
```diff
@@ -0,0 +1,113 @@
...
+  fi
+
+  # Drop the database if we need to force the install every time.
+  if [ "$FORCE_DATABASE_DROP" == 'true' ]; then
+    echo 'Dropping the Drupal DB if it exists'
+    "drush sql-drop -y -r ${WEB_DIRECTORY}"
+  fi
+
+  # If we're supposed to install Drupal, and it's not currently installed - then install it.
+  if ! drush status bootstrap | grep -q Successful ; then
```
kierenevans: Ditto here for the drush command (run it first, then check the exit code)

drupal8-apache/7.0/usr/local/share/container/baseimage-30.sh
```diff
@@ -1,41 +1,51 @@
...
-do_drupal8_development_start() {
-  bash /usr/local/share/drupal8/development/install.sh
-  do_drupal8_setup
-}
+#####
+# Tasks here are executed once all other services are available, so it should
+# be safe to to any tasks requiring databases, SOLR, etc.
+#####
+# alias_function do_setup do_drupal_setup_inner
+# do_setup() {
```
kierenevans: The `do_drupal_development_install` function would be perfect in this method, which you can then run with the following in your continuous-pipe.yml, after the database deployment:
```setup:
        run:
            image:
                from_service: web

            commands:
              - container setup
```

This also means it wouldn't try installing the site in each web container when INSTALL_DRUPAL is set to true and more than one web container is wanted spinning up at once.

andytson:
I'm not sure moving the "sensitive" permissions to development is necessary. It's already under a check that only applies to development mounts

drupal8-apache/7.0/usr/local/share/drupal/drupal8_functions.sh
```diff
@@ -45,9 +45,6 @@ do_drupal_create_directories() {
 #####
 do_drupal_permissions() {
   if [ "$IS_CHOWN_FORBIDDEN" -ne 0 ]; then
-    # Give the docroot and config directories to the web user.
-    chown -R "${APP_USER}":"${APP_GROUP}" "${WEB_DIRECTORY}"
-
```
andytson: I don't think this should be moved

drupal8-apache/7.0/Dockerfile
```diff
@@ -8,7 +8,6 @@ RUN curl -q https://dx6pc3giz7k1r.cloudfront.net/GPG-KEY-inviqa-tools | apt-key
  && bash /tmp/install-node.sh \
  && apt-get update -qq -y \
  && DEBIAN_FRONTEND=noninteractive apt-get -qq -y --no-install-recommends install \
-    hem \
```
andytson-inviqa: @kierenevans what do you think of this? any impact?

kierenevans: No drupal projects currently use hem, so no concerns from me. There's more that can be removed though

PR #179, opened 2017-02-28
Fix Firefox taking a long time to verify the chain

kierenevans:
For self-signed certificates. Similar fix to https://github.com/inviqa/chef-magento/pull/79

Needs to be tested! Seems to only affect OSX. Windows 10 unaffected by the slow resolution time.

### DIFF
--- a/nginx/usr/local/share/nginx/nginx_functions.sh
+++ b/nginx/usr/local/share/nginx/nginx_functions.sh
@@ -21,6 +21,7 @@ function do_https_certificates() {
             -days 365 \
             -newkey rsa:2048 \
             -keyout "${WEB_SSL_PRIVKEY}" \
+            -outform PEM \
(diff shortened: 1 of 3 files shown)


kierenevans:
This doesn't fix the issue.

It looks like the long delay in connecting over SSL with a self signed certificate in Firefox is due to https://bugzilla.mozilla.org/show_bug.cgi?id=1056341 (some extra details on https://bugzilla.redhat.com/show_bug.cgi?id=1204670 )

The problem has been noticed more on drupal and magento installs which don't really set WEB_HOST correctly. Magento sets APP_HOSTNAME by default and drupal doesn't set anything.

I guess the re-generation of certificates on each launch of the container could lead to the same build-up in Firefox's certificate store even for the same hostname in the CommonName field.
A suggested workaround is putting different values in the DistinguishedName field, which we aren't setting here.

PR #180, opened 2017-03-03
Fix magento2 cron

kierenevans:
Fix magento2 cron by having the full path to the command being executed in supervisor's config.


kierenevans:
Tested and (naughtily) already in the stable version after being tested.

PR #181, opened 2017-03-06
Standardise return values

kierenevans:
Switch from 0/1 to 'true'/'false'. Requires testing.

Might be a better way (comparison to /bin/true or /bin/false) but this at least helps us override the variables easily from docker-compose.yml and continuous-pipe.yml, which are text based.

We need to consider the BC aspect of this too, as a lot of deployments have "DEVELOPMENT_MODE" being "0" in the continuous-pipe.yml or docker-compose.yml. We could try to interpret "0" as true and "1" as false?


sroze:
I believe that we can convert this BC quite easily in the projects we are aware of. For those we don't know about, I suggest we can consider that they will have to update their configuration as this is a very logical change on our end that, IMHO, justify the BC break.

ubuntu/16.04/usr/local/share/container/baseimage-10.sh
```diff
@@ -23,7 +23,7 @@ do_build() {
 }
 
 check_development_start() {
-  if [ "$DEVELOPMENT_MODE" -eq 0 ]; then
+  if [ "$DEVELOPMENT_MODE" == 'true' ]; then
```
sroze: Does that mean that the value need to be the `true` string and can't be the integer `1` for example? As I don't see any call to convert this value?

kierenevans: Yes that's right, though can be changed.

sroze: That would be nice if you could change it as I can see some troubles coming from YAML reading `true` as a boolean, etc...

kierenevans: Conversions from `0`/`1` to `true`/`false` are now in place. What does YAML pass through boolean `true`/`false` as?

sroze: YAML have booleans. But I think that it's converted as `0` & `1` when injected through environment variables by k8s.

kierenevans:
This has been rebased on top of latest master.

kierenevans:
I have just found a bug with our approach:
In confd, `{{ if getenv "AUTH_HTTP_ENABLED" }}{{ getenv "AUTH_HTTP_ENABLED" }}{{ end }}` succeeds when it should fail. AUTH_HTTP_ENABLED prints as `false`

PR #182, opened 2017-03-07
Create a ssh-forward image for supporting local/remote port forwarding and …

andytson-inviqa:
…SOCKS proxy

This is multi-purpose for example to solve:

* Seeing behat tests running in a local browser
* Using PHP XDebug remote debugging in an IDE
* SOCKS proxying into Docker for Mac for direct browsing unexposed http/https

### DIFF
--- a/ssh-forward/usr/local/share/container/baseimage-20.sh
+++ b/ssh-forward/usr/local/share/container/baseimage-20.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+do_ssh_forward_set_credentials() {
+  if [ -n "${SSH_FORWARD_PASSWORD}" ]; then
+    echo "forward:${SSH_FORWARD_PASSWORD}" | chpasswd
+  fi
+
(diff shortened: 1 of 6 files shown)


ssh-forward/usr/local/share/container/baseimage-20.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+do_ssh_forward_set_credentials() {
+  if [ -n "${SSH_FORWARD_PASSWORD}" ]; then
+    echo "forward:${SSH_FORWARD_PASSWORD} | chpasswd
```
kierenevans: Missing quote mark

andytson-inviqa: fixed

kierenevans:
Do we need fail2ban in here as well?

andytson-inviqa:
I would suggest the port is never exposed publicly

PR #185, opened 2017-03-12
Allow HTTP by default

sroze:
Disallowing HTTP by default is a very bad user experience. In most of the infrastructure, the HTTPS termination is done at the proxy gateway layer rather than in the container. Using HTTPS between containers is a rare use case, that should be activated via an opt-in.


andytson-inviqa:
don't allow http and https just turn off https and http will "auto" be true. We actually still need to adjust this more for https proxy, as the redirect to https should still happen.

A proxied backend will NEED the reverse proxy and https offload setting to be set to true, so I don't think this PR should solve it by turning http on, instead the reverse proxy should be opt-in, rather than messing with http/https defaults

andytson-inviqa:
This in addition to local docker not needing reverse proxy, so should default to https

sroze:
So what is the real benefit of this "mandatory HTTPS" by default? We need to measure the impact it has on actual users in various environments, that's is the most important thing. This is about usability here, not anything else.

I can tell you from my experiences and from other people that told me, that this is painful for them to have that strict setup by default. They need to understand that they need to play with this environment variable (`WEB_HTTP`) plus actually this one (the one that says the SSL is offloaded somewhere else) oh plus the other one (that tells that it installs the dev dependencies so they can see what's wrong in their app) to deploy their container successfully...

My point of view is that this on-boarding is the *most important*. The rest, which is about security and optimisations erm deploying to production is something that needs to happen but first it needs to work.

andytson:
this is a purely GKE issue, not a docker issue though. Httplabs could talk to the container via https

andytson:
people should read usage, and their not going to understand GKE https ingress configuration unless they read it from somewhere. That somewhere could say what is required of the container it connects to.

andytson:
I'll support a single env var that should be set to turn it into a reverse proxy backend. But in reality the backend also needs to know much more than that, such as trusted proxy IP's, and IP whitelists

kierenevans:
#183 handled the x-forwarded-proto redirection for the base nginx image. Still need to apply to php-nginx.

sroze:
The background being that I've ran an experiment with the nginx ingress controller (in front of just one GCE ingress) and that's working perfectly. See here for all the configuration we can do on the nginx ingresses, we probably won't have to do that much (if not nothing) on the Docker images:
https://github.com/kubernetes/ingress/blob/master/controllers/nginx/configuration.md

kierenevans:
Similar to developing with varnish in the stack, if sites are going HTTPS-only to help stop interception on the wire and get a boost in the Google search results, we should develop that way too to pick up issues before they hit production.

These images were aimed at not just cp but docker in general, including local development which wouldn't have the kubenetes ingress available.

Reducing the amount of variables to configure would be a good move but without 301ing to HTTPS, most people will not bother with HTTPS at all. This also means people won't benefit from http2 and will lead users to focus on obsolete performance improvements like domain sharding.

kierenevans:
I would also argue that container to container communications should be encrypted, in an ideal world.
I understand this isn't feasible in a lot of cases.

andytson-inviqa:
#205 is a compromise to make it more clear what's doing the redirect and how to turn it off

PR #188, opened 2017-03-16
NGINX: Look up the IP of the hostname

kierenevans:
NGINX's mod_realip needs an IP address 
If a hostname is provided, or even if it's an IP address, look up the host with `getent hosts` and parse the value.
This allows setting up a DNS hostname for a load balancer that sits in front of this container.


andytson-inviqa:
is this stuff needing to be copied to the php-nginx tempalte

nginx/etc/confd/templates/nginx/nginx_ip_whitelist_realip.conf.tmpl
```diff
@@ -1,7 +1,7 @@
-{{ if getenv "EXTERNAL_LOAD_BALANCER_IP" }}
+{{ if getenv "EXTERNAL_LOAD_BALANCER_HOST_IPS" }}
```
andytson-inviqa: TRUSTED_REVERSE_PROXIES should be allowed irrespective of whether there's an endpoint

kierenevans: Makes sense, updated to use an "or"

kierenevans:
> is this stuff needing to be copied to the php-nginx templates
Yes, though the previous PR #183 hasn't been yet either due to lack of time.

PR #189, opened 2017-03-17
Copy nginx image realip/healthcheck/reverse proxy https redirects cha…

andytson-inviqa:
…nges to php-nginx

Also nginx_ip_whitelist_realip renamed to nginx_realip due to being used also in other situations (such as nginx access logs, and fraud checking solutions)


php-nginx/etc/confd/conf.d/nginx_realip.conf.toml
```diff
@@ -0,0 +1,6 @@
+[template]
+src   = "nginx/realip.conf.tmpl"
```
kierenevans: `nginx/nginx_realip.conf.tmpl`

nginx/etc/confd/conf.d/nginx_site_redirect_to_https.conf.toml
```diff
@@ -1,6 +1,6 @@
 [template]
 src   = "nginx/site_redirect_to_https.conf.tmpl"
-dest  = "/etc/nginx/sites-available/default-20-site_redirect_to_https.conf"
+dest  = "/etc/nginx/sites-available/default-05-site_redirect_to_https.conf"
```
kierenevans: It was moved from 05 to 20 in https://github.com/continuouspipe/dockerfiles/commit/48511980ea511737486ba76b782eb2fa8068bc0b, and it might still be needed to stop giving a 301 to a healthcheck.

andytson-inviqa: ok, it's a bit inconsistent then, with php-nginx it being after the first location {} block that's in site_base. I'll move the healthcheck to before it too. The root only needed for things that resolve to actual files

php-nginx/etc/confd/conf.d/nginx_realip.conf.toml
andytson-inviqa: fixed

nginx/etc/confd/conf.d/nginx_site_redirect_to_https.conf.toml
andytson-inviqa: fixed

PR #191, opened 2017-03-20
Improvements to mongodb users update and connection attempts

andytson-inviqa:


kierenevans:
I think the test failure is due to an update of Shellcheck. You can cherry-pick https://github.com/continuouspipe/dockerfiles/commit/35823c284d920ebb217dc6e715268de21d0dd565 or at least apply the same fix, as it makes more sense in your PR (or another one)

andytson-inviqa:
I think quotes will break the declare, so I'll mark it ignored

PR #196, opened 2017-03-20
Add SSL server protocol and cipher configuration to Nginx/Apache images

andytson-inviqa:
The intention for the SSL protocols is to have TLS 1.0+ enabled
Nginx's protocols are secure and can't be negated, so if not specified Nginx's own default will apply
Apache's protocols aren't always secure, but can be negated, so all except SSLv2/3 set to allowed

The SSL configuration is now separated as an include so projects potentially overriding the overall vhost/server can still include the settings without duplicating them.


andytson-inviqa:
fixed

andytson-inviqa:
ah, lol, if a PR written by me is left unmerged for too long, I go and implement it again forgetting the original PR, so this duplicates #324

The #324 PR has been tested and includes more functionality, however, so I'll close this one

sroze:
😄

PR #198, opened 2017-03-20
Add Spryker image using Symfony-nginx as base

hgajjar-sessiondigital:


spryker/nginx/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM quay.io/continuouspipe/symfony-php7.1-nginx:latest
+
+MAINTAINER Hardik Gajjar <hardik.gajjar@inviqa.com>
+
+RUN apt-get update && \
+    apt-get install -y php7.1-pgsql php7.1-gmp postgresql-client graphviz redis-tools
```
kierenevans: Don't forget to clean up the apt caches after this.

Try to combine the two RUN commands into one, a good example is: https://github.com/continuouspipe/dockerfiles/blob/master/drupal8-apache/7.0/Dockerfile#L5-L15

spryker/nginx/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -0,0 +1,19 @@
...
+do_spryker_directory_create() {
+  mkdir -p /app/data/DE/cache/Yves/twig
+  mkdir -p /app/data/DE/cache/Zed/twig
+  mkdir -p /app/data/DE/logs
+  mkdir -p /app/data/common
+}
+
+do_spryker_config_create() {
+  # create .pgpass in home directory for postgres client
+  echo "$DATABASE_HOST:*:*:$DATABASE_USER:$DATABASE_PASSWORD" > ~/.pgpass
```
kierenevans: Should this be in the root user's home directory?

hgajjar-sessiondigital: Erm.. probably not, as the install command is going to be executed by code owner (I guess), so probably I should use "as_code_owner" fn.

PR #203, opened 2017-03-21
Append GatewayPorts yes to sshd config

blongden:


sroze:
Can you quickly describe why in this PR?

blongden:
Sure.

By default, remote port forwarding will bind to localhost - so only the server is able to connect to the port (ie, remote forwarding of local port 9000 will bind to localhost:9000 on the container). Setting GatewayPorts yes will bind the port forwarding to *:9000 to allow other containers to also connect to the forwarded port (ie, the php xdebug extension).

PR #204, opened 2017-03-22
Reverse proxy and https offload fixes

andytson-inviqa:
This is a compromise over #199.

Instead, WEB_REVERSE_PROXIED can be safely turned on for all our CP environments without dictating that for other users. We can separately consider whether the images should have it enabled by default.

WEB_HTTPS_OFFLOAD = use server port to be scheme authority, when https port doesn't use SSL.
WEB_REVERSE_PROXIED = use request headers if available to be scheme authority, otherwise fall back to nginx scheme

### DIFF
--- a/nginx/etc/confd/templates/nginx/site_custom_scheme_flags.conf.tmpl
+++ b/nginx/etc/confd/templates/nginx/site_custom_scheme_flags.conf.tmpl
@@ -1,4 +1,3 @@
-{{ if eq "false" (getenv "WEB_REVERSE_PROXIED") }}
 set $custom_https $https;
 set $custom_scheme $scheme;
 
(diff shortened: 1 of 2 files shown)


andytson-inviqa:
client http -> reverse proxy (port matching) -> container http:80
client https -> reverse proxy (port matching) -> container https:443

client http -> reverse proxy (port matching) -> container(WEB_REVERSE_PROXIED) http:80
client https -> reverse proxy (port matching) -> container(WEB_REVERSE_PROXIED) https:443

client http -> reverse proxy (port matching and offloading) -> container(WEB_HTTPS_OFFLOAD) http:80
client https -> reverse proxy (port matching and offloading) -> container(WEB_HTTPS_OFFLOAD) http:443

client http -> reverse proxy (X-Forwarded-Proto matching) -> container(WEB_REVERSE_PROXIED,WEB_HTTP) http:80
or
client http -> reverse proxy (X-Forwarded-Proto matching) -> container(WEB_REVERSE_PROXIED) https:443

No reason to use WEB_HTTPS_OFFLOAD and WEB_REVERSE_PROXIED at the same time.

PR #208, opened 2017-03-22
Copy behaviour for WEB_REVERSE_PROXIED and WEB_HTTPS_ONLY to apache image

andytson-inviqa:


andytson-inviqa:
I've tested this locally, and it appears to behave as intended

andytson-inviqa:
well, aside from being able to fool PHP into thinking the following is http:
`curl -H 'X-Forwarded-Proto: http' -k https://localhost/`

This is due to Apache overwriting the HTTPS env var with it's true value

PR #209, opened 2017-03-23
Error and quit if an unset variable is used in functions.

kierenevans:
Helps avoid `rm -rf ${VAR}/*` wiping /* if $VAR is not set, as it could wipe out mountpoints.

A few variables such as GITHUB_TOKEN are impacted, this PR will gradually build over time to ensure variables are declared before being used.

### DIFF
--- a/ubuntu/16.04/usr/local/bin/container
+++ b/ubuntu/16.04/usr/local/bin/container
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -e -u
 
 TASK=$1
 ARGS=( "$@" )


sroze:
> this PR will gradually build over time to ensure variables are declared before being used

What do you mean?

kierenevans:
GITHUB_TOKEN is currently not set anywhere, but we check if it's set with `if [ -n "${GITHUB_TOKEN}" ]`.

That would not work with this -u flag and would exit the script. We should declare that and other variables in env files before use.

PR #210, opened 2017-03-28
Kill supervisord when managed processes enter FATAL state

kierenevans:
Kill the supervisord process with a SIGTERM if a process it manages enters the fatal state, which means it didn't manage to start up successfully at all.

Tested with the following conf file `ubuntu/16.04/etc/supervisor/conf.d/test.conf`:
```
[program:test]
command = /bin/false
stdout_logfile=/dev/stdout
stdout_logfile_maxbytes=0
stderr_logfile=/dev/stderr
stderr_logfile_maxbytes=0
loglevel = warn
user = root
autostart = true
autorestart = true
priority = 5
```

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -11,7 +11,7 @@ fi
 docker pull koalaman/shellcheck
 docker pull lukasmartinelli/hadolint
 
-find "$DIR" -type f ! -path "*.git/*" \( \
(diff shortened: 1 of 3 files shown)


sroze:
🎉 🍻

PR #211, opened 2017-03-30
Remove deprecated antelope tool

blongden:


spryker/nginx/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -27,9 +27,8 @@ do_spryker_build() {
 }
 
 do_build_assets() {
-  as_code_owner "cd /app && antelope install"
-  as_code_owner "cd /app && antelope build zed"
-  as_code_owner "cd /app && antelope build yves"
+  as_code_owner "cd /app && npm run zed"
```
kierenevans: Default run location is /app, so no need to cd to it

spryker/nginx/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -27,9 +27,8 @@ do_spryker_build() {
...
+  as_code_owner "cd /app && npm run yves"
```
kierenevans: Default run location is /app, so no need to cd to it

PR #212, opened 2017-03-30
Add an function allowing do_start to be moved to an entrypoint

andytson-inviqa:
This is useful for CLI apps

### DIFF
--- a/ubuntu/16.04/Dockerfile
+++ b/ubuntu/16.04/Dockerfile
@@ -43,5 +43,4 @@ RUN echo 'APT::Install-Recommends 0;' >> /etc/apt/apt.conf.d/01norecommends \
 COPY ./etc/ /etc/
 COPY ./usr/ /usr/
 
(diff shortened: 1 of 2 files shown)


andytson-inviqa:
seems to work nicely:
```
docker run -it --rm quay.io/continuouspipe/ubuntu16.04
# runs do_start then do_supervisord, supervisord becomes pid 1

docker run -it --rm quay.io/continuouspipe/ubuntu16.04 bash, bash becomes pid 1
# runs do_start then opens a bash tty
```

andytson-inviqa:
can override the behaviour using:
```
docker run -it --rm --entrypoint '' quay.io/continuouspipe/ubuntu16.04 bash
```

andytson-inviqa:
I'm not sure it'll be "fully" BC, or what CP does with it in run tasks, so I think that should be tested first

andytson-inviqa:
I'll make it an optional thing instead I think

PR #216, opened 2017-04-04
Run Spryker setup command only if its not already installed

hgajjar-sessiondigital:


blongden:
Looks good! let's get it merged :)

PR #219, opened 2017-04-05
Add Slack notification on failure

kierenevans:


sroze:
@kierenevans did you configure the variable?

kierenevans:
Yep!

PR #224, opened 2017-04-09
Improve Magento 2 Varnish VCL

### LINKED ISSUE #155: Magento2 Varnish VCL is in development mode
https://github.com/continuouspipe/dockerfiles/blob/4d9296030217fdad0c77c93ad3d4a7f6d7628d0a/magento2-varnish/4.0/etc/confd/templates/varnish/default.vcl.tmpl#L66-L69 should be made optional, as if used in a non-development environment, it will be not caching things it should.

kierenevans:
* Allow MAGENTO_MODE to be passed in. 
* Support additional IP addresses for PURGE

Fixes #155


sroze:
Looks good to me 👍

PR #225, opened 2017-04-10
add piwik

alexbiddle:
Add piwik installation to CP

Test with composer file (not linked to db at moment, you can just see the installation screen)

````version: '2'
services:
  web:
    image: quay.io/continuouspipe/piwik-php7.1-apache:latest
    ports:
    - '443:443'

### DIFF
--- a/piwik/Dockerfile-php7.1
+++ b/piwik/Dockerfile-php7.1
@@ -1,0 +1,13 @@
+FROM quay.io/continuouspipe/php7.1-apache:latest
+
+MAINTAINER Alex Biddle <alex.biddle+cp-dockerfiles@inviqa.com>
+
+WORKDIR /app/web/piwik
+RUN curl -L -O https://builds.piwik.org/latest.tar.gz && \
+    tar --strip 1 -xzf latest.tar.gz && \
+    rm latest.tar.gz
+
+RUN chown -R build:www-data ./*
+RUN chmod -R 0755 ./*
+RUN chmod -R 0775 tmp \
+    config
--- a/piwik/README.md
+++ b/piwik/README.md
@@ -1,0 +1,31 @@
+# Piwik
+
+In a docker-compose.yml:
+```yml
+version: '3'
+services:
+  piwik:
+    image: quay.io/continuouspipe/piwik-php7.1-apache:stable
+```
+
+In a Dockerfile:
+```Dockerfile
+FROM quay.io/continuouspipe/piwik-php7.1-apache:stable
+```
+
+## How to build
+```bash
+docker-compose build piwik_php71_apache
+docker-compose push piwik_php71_apache
+```
+
+## About
+
+This is an image that installs [piwik](http://piwik.org) web analytics software.
+
+
+### Environment variables
+
+No additional environment variables over those of continuouspipe/php7.1-apache are supported, as piwik at this time does not support headless installations.
+
+
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ internal services and for the application templates.
     - [NGINX w/ PHP-FPM 7.0](magento2-nginx/fpm-7.0/)
     - [Varnish](magento2-varnish/4.0/)
 - [Symfony - NGINX or Apache with PHP 5.6, 7.0 or 7.1)](symfony/)
+- [Piwik](piwik/)
 
 ### Web Applications
 
--- a/continuous-pipe.yml
+++ b/continuous-pipe.yml
@@ -171,6 +171,9 @@ tasks:
         phantomjs2:
           image: quay.io/continuouspipe/phantomjs2
           tag: latest
+        piwik_php71_apache:
+          image: quay.io/continuouspipe/piwik-php7.1-apache
+          tag: latest
         redis:
           image: quay.io/continuouspipe/redis3
           tag: latest
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -191,6 +191,14 @@ services:
     depends_on:
       - ubuntu
 
+  piwik_php71_apache:
+    build:
+      context: ./piwik/
+      dockerfile: Dockerfile-php7.1
+    image: quay.io/continuouspipe/piwik-php7.1-apache:latest
+    depends_on:
+      - php71_apache
+
   php71_apache:
     build:
       context: ./php-apache/

### CODE REVIEW

**Comment by kierenevans** [comment]
Hi @alexbiddle, thank you for your PR.

Can you take a look at the Travis feedback from the running of the `hadolint` tool, which comments on best practises for Dockerfiles.

The following steps will also be needed to automatically build your new image with CP when merged to master. We should probably document them, and are available to assist you:
1. Add an entry into https://github.com/continuouspipe/dockerfiles/blob/master/docker-compose.yml that says to build this directory, declaring the dependency on php70_apache
2. Add a corresponding entry for the docker-compose service name into https://github.com/continuouspipe/dockerfiles/blob/master/continuous-pipe.yml in the right place to pick up on the dependency of the php70_apache image.
3. Document the image in the same style as other README.md's (the symfony one is one of the best so far)
4. Create a quay.io repository under the continuouspipe namespace that the CP robot account can push to (we can assist with this once you've decided on a name)

Thanks,
Kieren

File: piwik/etc/confd/templates/virtual-host.conf.tmpl [hunk 1]
```diff
@@ -0,0 +1,17 @@
+<VirtualHost *:80>
```
kierenevans: This file may not be needed, as the templates in https://github.com/continuouspipe/dockerfiles/tree/master/php-apache/etc/confd/templates/apache should handle this.

piwik/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+RUN echo "extension=geoip.so" >> /etc/php/7.0/apache2/php.ini
+RUN echo "geoip.custom_directory=/app/misc" >> /etc/php/7.0/apache2/php.ini
+
+WORKDIR /app/web
+RUN curl -L -O https://builds.piwik.org/latest.tar.gz && \
+    tar --strip 1 -xzf latest.tar.gz && \
+    rm latest.tar.gz
+
+RUN chown -R www-data:www-data *
```
kierenevans: This isn't the safest of operations, are you sure that the web user need to own all of /app/web to function? The safest way to solve this would be to run the untar command above as the `build` user, then reset the permissions for individual web writable files/folders if they must be.

piwik/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM quay.io/continuouspipe/php7-apache
+
+MAINTAINER Alex Biddle <alex.biddle+cp-dockerfiles@inviqa.com>
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y php-geoip php-dev libgeoip-dev && apt-get clean
+
+RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+RUN echo "extension=geoip.so" >> /etc/php/7.0/apache2/php.ini
```
kierenevans: A confd template that writes to `/etc/php/7.0/apache2/conf.d/geoip.ini` would be a more user friendly way to enable the geoip module, as users who use this image could control that template as a whole.

piwik/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    rm latest.tar.gz
+
+RUN chown -R www-data:www-data *
+RUN chmod -R 0755 tmp
+
+
+WORKDIR /app/misc
+RUN curl -L -O http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz && gunzip GeoIP.dat.gz
+RUN curl -L -O http://geolite.maxmind.com/download/geoip/database/GeoLiteCity.dat.gz && gunzip GeoLiteCity.dat.gz && mv GeoLiteCity.dat GeoIPCity.dat
+RUN chmod a+w /app/misc
```
kierenevans: This would be better as just being readable by the web server user, rather than writable too unless piwik has a built in auto-update mechanism for GeoIP.
Perhaps the geoip package provided by ubuntu and corresponding cronjob would be a better choice?

piwik/etc/confd/templates/virtual-host.conf.tmpl
andytson-inviqa: it isn't needed, and shouldn't be used. There doesn't seem to be a reason for adding the /server-status, but if there is, then please read https://github.com/continuouspipe/dockerfiles/tree/master/php-apache#addingreplacing-apache-httpd-configuration.

Piwik, like Google Analytics does, should be used with HTTPS only, so inheriting the base image template would enable that

piwik/Dockerfile
andytson-inviqa: I don't think a confd template is needed. If a user wants to configure some geoip settings (if there are any), then they could just add their own custom file (as with any other extension)

However, adding to php.ini I don't recommend. Instead it should use the conf.d folder for dynamic insertion.

alexbiddle:
Hi everyone, thanks for your comments. I'll get the changes made and tested an push up again later.

piwik/etc/confd/templates/virtual-host.conf.tmpl
alexbiddle: Removed

piwik/Dockerfile
alexbiddle: Actually, the need for geoip functionality is out of scope for this task, so I'm going to just remove this.

alexbiddle:
I have made the changes, I used the magento README file as a template--I hope it's OK.

Running 
    docker-compose build piwik_php71_apache

builds fine

piwik/README.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+## About
+
+This is an image that installs [piwik](http://piwik.org) web analytics software.
+
+
+### Environment variables
+
+No environmental variables are supported, as piwik at this time does not support headless installation.
```
andytson-inviqa: No additional environment variables over those of continuouspipe/php7.1-apache are supported.

piwik/README.md
```diff
@@ -0,0 +1,31 @@
+# Piwik
+
+In a docker-compose.yml:
+```yml
+version: '3'
+services:
+  piwik:
+    image: quay.io/continuouspipe/piwik-php7.1-apache:latest
```
andytson-inviqa: we use `:stable` in the documentation, as we take cuts from `:latest` as stable tags when we're happy with them.

piwik/README.md
alexbiddle: will change

piwik/README.md
alexbiddle: done

piwik/README.md
```diff
@@ -0,0 +1,31 @@
...
+```yml
+version: '3'
+services:
+  piwik:
+    image: quay.io/continuouspipe/piwik-php7.1-apache:stable
+```
+
+In a Dockerfile:
+```Dockerfile
+FROM quay.io/continuouspipe/piwik-php7.1-apache:latest
```
andytson-inviqa: and here

alexbiddle:
@kierenevans 

I think my latest commit addresses all the issues, apart from 4. I decided on the name quay.io/continuouspipe/piwik-php7.1-apache, following the magento2 images.

kierenevans:
Ready for merging when https://quay.io/repository/continuouspipe/piwik-php7.1-apache starts working :)

alexbiddle:
Cool, thank you 👍

piwik/Dockerfile-php7.1
```diff
@@ -0,0 +1,13 @@
+FROM quay.io/continuouspipe/php7.1-apache:latest
+
+MAINTAINER Alex Biddle <alex.biddle+cp-dockerfiles@inviqa.com>
+
+WORKDIR /app/web/piwik
```
alexbiddle: installed into its own directory

alexbiddle:
So I moved the piwik installation into its own directory, which now works great for my purposes.

How is quay going? 

Cheers, Alex

kierenevans:
It's either dashes or dots that are the problem, so I've created https://quay.io/repository/continuouspipe/piwik as a temporary measure. Is it okay to update continuous-pipe.yml and the README.md to reference that?

kierenevans:
Just waiting on Quay support though I have managed to successfully push `quay.io/continuouspipe/piwik:latest` to unblock you :)

No need to update this PR to match that (unlike my previous comment said), as I couldn't create a robot user to push to the piwik repo either!

kierenevans:
So, erm, "piwik" is of course a tracking technology, as such will be blocked by adblockers such as ublock origin. This includes on quay.io, which is why it was not working for me.

PR #226, opened 2017-04-10
Set MAGE_MODE too, so the magento2 cli picks up on the magento mode being set from nginx.

kierenevans:
Magento 2's tooling picks up MAGE_MODE. We set MAGENTO_MODE.


adragus-inviqa:
I'm not sure this is worth it, or even working. [M2 suggests setting the mode via its CLI tool](http://devdocs.magento.com/guides/v2.1/config-guide/bootstrap/magento-how-to-set.html).

Also, check this link, too: http://devdocs.magento.com/guides/v2.0/config-guide/cli/config-cli-subcommands-mode.html

I think they want us to use the CLI tool, because a transition between modes needs other operations, like compiling, temporarely entering maintenance mode etc. A simple env var change wouldn't be able to do that.

kierenevans:
I believe this PR was to pull the CLI tool in line with what we do with the NGINX template with MAGENTO_MODE: https://github.com/continuouspipe/dockerfiles/blob/master/magento2-nginx/fpm-7.0/etc/confd/templates/nginx/site_base.conf.tmpl#L2

The CLI only picks up on env.php and MAGE_MODE env variables, so this was something quick to get the mode being reported accurately.

This might be a better change perhaps, though there are some projects that override this template: https://github.com/continuouspipe/dockerfiles/blob/master/magento2-nginx/fpm-7.0/etc/confd/templates/magento/env.php.tmpl#L92

PR #234, opened 2017-04-13
Enable headers and expires modules for Drupal .htaccess use.

kierenevans:
Lots of IfModule checks in the .htaccess that drupal 8 ships with.


andytson-inviqa:
I'd prefer these enabled in the php-apache image I think

andytson-inviqa:
nvm, since it's not an install but a config enable it doesn't matter too much

PR #238, opened 2017-04-25
Build the NGINX Ingress Controller image

kierenevans:
Imported from https://github.com/sroze/kubernetes-experiments/pull/6

Also fixes a cp.yml duplicate task bug


sroze:
I would prefer not building it and recommending the official image 🤔

kierenevans:
I would really like this too. For now we could build it and hopefully a newer version will do what we need it to do with the SSL support?

sroze:
Did we contribute back to the repository? If not, could we easily do so?

PR #242, opened 2017-05-03
Use hem to apply assets in magento1-apache-php5.6 image

jameshalsall:


kierenevans:
Hi @jameshalsall, thank you for your PR. We were considering removing hem altogether from these images to save ~500MB per project, for example using the `awscli` here: https://github.com/continuouspipe/dockerfiles/blob/e9e901b3a1809c662465a4c06c3c523818a35929/magento2-nginx/fpm-7.0/usr/local/share/magento2/magento_functions.sh#L109-L119

Does the tar not do something that you wish it to?

andytson-inviqa:
@kierenevans it was mostly to solve an issue that ASSET_ARCHIVE_PATH was relative to project root but the command was cd'd into public/media. There are a lot of assumptions in the code as well, which Hem took care of, so it seemed best to replace with Hem, as it was known to work as previously outside docker

jameshalsall:
Let me know if you think this needs re-working to use the awscli @kierenevans

PR #246, opened 2017-05-04
Magento 1 image fixes

andytson-inviqa:
mysql-client wasn't present in image, causing n98-magerun to fail
Apache config wasn't properly handling index.php rewrite (pushed to php-apache image as a general fix to match php-nginx)
API rewrite is useful


andytson-inviqa:
travis push passed btw, just it hasn't informed github

PR #252, opened 2017-05-11
Fix drupal README formatting

alexbiddle:
Minor fixes


sroze:
Thank you @alexbiddle

PR #254, opened 2017-05-14
Add make to base image

sroze:


andytson-inviqa:
@kierenevans and I have discussed making the plan.sh more accessible and documented to try to combat the issues of not having a standard way of adding tasks that can be ported easily to the base images

PR #259, opened 2017-05-16
Deprecate mysql root user usage and grants

andytson-inviqa:
The MySQL application user is better set up via the mysql image's environment variables.


andytson-inviqa:
closing in favour of #260

PR #264, opened 2017-05-23
Highly available redis3 with customisable master failover timeout

kierenevans:
Based on a Google image built from alpine.

Image source is available here: https://github.com/kubernetes/kubernetes/blob/master/examples/storage/redis/image/

We were seeing failovers in a redis cluster take 60 seconds to detect, which is the setting in the run.sh from the kubernetes/kubernetes link above.
We have therefore allowed customisation and defaulted to a lower amount of time as per the tutorial on https://redis.io/topics/sentinel#a-quick-tutorial

### DIFF
--- a/continuous-pipe.yml
+++ b/continuous-pipe.yml
@@ -183,6 +183,9 @@ tasks:
         redis:
           image: quay.io/continuouspipe/redis3
           tag: latest
+        redis_highly_available:
+          image: quay.io/continuouspipe/redis3-highly-available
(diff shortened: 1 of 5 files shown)


kierenevans:
Had to update dpkg and openldap manually as they weren't in the security updates but still flagged as problematic in quay's security scan.
Should we ignore https://github.com/lukasmartinelli/hadolint/wiki/DL3005 ?

PR #265, opened 2017-05-31
Use all X-Forwarded-* headers by default

andytson-inviqa:
Some applications (e.g. Symfony) implicitly trust other X-Forwarded-* headers if they are sent from a trusted proxy. This means they can be manipulated by the user if not set or cleared.

Therefore a new "edge_headers" location setting, defaulting to true, can be used to say which trusted proxy is considered the edge server, and a new "hide_edge_headers" location setting to unset any earlier trusted proxy headers, when it's known there are none or not supplied.

X-Forwarded-For isn't just an edge header but a full chain, so it should be set whatever the case.

Sadly this wont cater to every reverse proxy situation, but is the best fit for most.

### DIFF
--- a/nginx-reverse-proxy/README.md
+++ b/nginx-reverse-proxy/README.md
@@ -47,7 +47,31 @@ export PROXY_LOCATIONS='[
 Multiple locations can be specified. The only required keys per location are
 location and backend.
 
+Location settings available:
+
+Key | Description | Expected values | Default
+--- | --- | --- | ---
+location | The domain-relative uri for which to apply the proxy configuration | domain-relative uri |
+backend | The backend host to proxy for the location, note for domains, WEB_RESOLVER needs to be set | url |
+preserve_host | Whether to pass the request's HTTP Host header value, or otherwise use the host from the backend setting | true/false | false
+use_downstream_edge_headers | Whether to pass the downstream edge headers to the backend server, or send it's own | true/false | false
+hide_edge_headers | Whether to hide edge headers from the backend server | true/false | false
+
 As this is using NGINX's proxy module, check out the documentation here: https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass
+
+### Edge vs intermediate reverse proxy header handling
+
+Edge and intermediate reverse proxies should handle X-Forwarded-* headers differently.
+
+Other than X-Forwarded-For, generally edge servers should ignore downstream
+X-Forwarded-* headers as they are untrusted, yet intermediate servers may need
+to proxy downstream X-Forwarded-* headers, if the edge servers set them.
+
+This can be achieved by setting `use_downstream_edge_headers` to true for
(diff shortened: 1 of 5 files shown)


sroze:
How WIP is that @andytson-inviqa ? :)

andytson-inviqa:
sadly this is failing because isset isn't in the version of go text/templates in use.

Confd appears to be abandoned as it's had no updates since Feb 2016. Kelsey changed jobs and this has prob become a less of a priority for him.

I think we should prob try to find another template tool, or something.

andytson-inviqa:
I've renamed the variable to suggest the opposite behaviour to solve it (with the default being false instead of true).

Tested and seems to work as expected.

nginx-reverse-proxy/README.md
```diff
@@ -47,8 +47,32 @@ export PROXY_LOCATIONS='[
...
 
+Location settings available:
+
+Key | Description | Expected values | Default
+--- | --- | --- | ---
+location | The domain-relative uri for which to apply the proxy configuration | domain-relative uri |
+backend | The backend host to proxy for the location, note for domains, WEB_RESOLVER needs to be set | url |
+preserve_host | Whether to pass the request's HTTP Host header value, or otherwise use the host from the backend setting | true/false | false
+use_downstream_edge_headers | Whether to pass the downstream edge headers to the backend server, or send it's own | true/false |
+hide_edge_headers | Whether to hide edge headers from the backend server | true/false |
```
sroze: Can you add the default value?

nginx-reverse-proxy/README.md
```diff
@@ -47,8 +47,32 @@ export PROXY_LOCATIONS='[
...
 location and backend.
 
+Location settings available:
+
+Key | Description | Expected values | Default
+--- | --- | --- | ---
+location | The domain-relative uri for which to apply the proxy configuration | domain-relative uri |
+backend | The backend host to proxy for the location, note for domains, WEB_RESOLVER needs to be set | url |
+preserve_host | Whether to pass the request's HTTP Host header value, or otherwise use the host from the backend setting | true/false | false
+use_downstream_edge_headers | Whether to pass the downstream edge headers to the backend server, or send it's own | true/false |
```
sroze: Can you add the default value?

nginx-reverse-proxy/etc/confd/templates/nginx/site_base.conf.tmpl
```diff
@@ -3,10 +3,20 @@
 location {{ .location }} {
   proxy_pass {{ .backend }};
   proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+{{ if .hide_edge_headers }}
```
sroze: How will the user set this value? 🤔

andytson-inviqa: already said in the readme, as are all the other options

PR #271, opened 2017-06-07
Added Yarn to Drupal images

Chapabu:
We've added Yarn to almost every project so far, so it makes sense to have it in the base image.


andytson-inviqa:
should we switch all base image calls to `npm` to `yarn` as well?

kierenevans:
There doesn't appear to be any frontend setup for the drupal image currently but it would be nice for the other images.

andytson-inviqa:
yarn doesn't appear to port over npm shrinkwrap files, so forcing migration might cause issues where dependencies are loose

andytson-inviqa:
This PR doesn't force in any case for drupal

Chapabu:
You can still use `npm` if `yarn` is installed, so that should be fine.

PR #285, opened 2017-06-13
Configure upstream varnish hosts in magento2 config

kierenevans:
The database settings `system/full_page_cache/varnish/access_list`, `system/full_page_cache/varnish/backend_host` and `system/full_page_cache/varnish/backend_port` only control how to generate a VCL from the Magento admin, they don't help us clear caches.

This configuration tells magento how to clear caches.


magento2-nginx/fpm-7.0/etc/confd/templates/magento/env.php.tmpl
```diff
@@ -110,4 +110,15 @@ return array (
         array (
             'date' => 'Tue, 24 Jan 2017 09:12:35 +0000',
         ),
+{{ if getenv "MAGENTO_HTTP_CACHE_HOSTS" }}
+    'http_cache_hosts' => array(
+{{ $varnish := split (replace (replace (getenv "MAGENTO_HTTP_CACHE_HOSTS") "," " " -1) "\n" " " -1) " " }}
```
sroze: I trust you but feel bad we can't test this 🤔

kierenevans: It's based on https://github.com/continuouspipe/dockerfiles/blob/master/nginx/etc/confd/templates/nginx/nginx_ip_whitelist.conf.tmpl#L2 if that helps any :)

sroze: 😛

PR #287, opened 2017-06-13
Set magento2 mode in the env.php

kierenevans:
This is in addition to #226 to allow magento2 to know it's mode from the CLI.

### DIFF
--- a/magento2-nginx/fpm-7.0/etc/confd/templates/magento/env.php.tmpl
+++ b/magento2-nginx/fpm-7.0/etc/confd/templates/magento/env.php.tmpl
@@ -89,7 +89,7 @@ return array (
                 ),
         ),
     'x-frame-options' => 'SAMEORIGIN',
-    'MAGE_MODE' => 'default',
+    'MAGE_MODE' => '{{ getenv "MAGENTO_MODE" }}',
     'cache_types' =>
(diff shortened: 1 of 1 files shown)


magento2-nginx/fpm-7.0/etc/confd/templates/magento/env.php.tmpl
```diff
@@ -89,7 +89,7 @@ return array (
                 ),
         ),
     'x-frame-options' => 'SAMEORIGIN',
-    'MAGE_MODE' => 'default',
+    'MAGE_MODE' => '{{ getenv "MAGENTO_MODE" }}',
```
sroze: Shouldn't we use `MAGE_MODE` ?

kierenevans: MAGENTO_MODE is still used in the NGINX config, MAGE_MODE was set to MAGENTO_MODE's value to let magento pick it up in the command line. This PR was intended to replace #226 so MAGE_MODE wouldn't have existed.

sroze: Oooh I see :)))

PR #288, opened 2017-06-13
Remove the need for basic auth to be on to return 200 to healthcheck

kierenevans:
For sites that aren't basic auth protected but still fronted by a load balancer that likes to check the health of the site a bit too often, return 200 immediately.


andytson-inviqa:
why `site_200_healthcheck` and not `site_healthcheck` btw?

kierenevans:
Not sure, tried to indicate it returns immediately with a 200. I don't mind renaming.

andytson-inviqa:
I think having 2 numbers in the end result file is a bit confusing

PR #290, opened 2017-06-13
Allow tuning opcache.

kierenevans:
Provide variables that allow tuning of opcache, such as whether to validate if files have been updated or not, and the size of cache allowed.


sroze:
Can you add these to the README? 🙏

kierenevans:
Added!

PR #291, opened 2017-06-13
Turn opcache.validate_timestamp to off in non-development mode

kierenevans:
PHP changes would need a reload of apache or php-fpm.
Depends on #290


andytson-inviqa:
This is one thing I love about docker, can turn this on knowing apps don't have a separate deployment process that will require invalidating the cache

andytson-inviqa:
Would be best mentioning it in the readme

PR #301, opened 2017-06-13
Output magento2 reports into the container logs

kierenevans:
Every minute, find the report files written out to /app/var/report/ and print them out to the container logs so they are discoverable.
Depends on #297 for the do_magento_echo_last_minute_reports function that is being called.


kierenevans:
The do_magento_echo_last_minute_reports function will be present in the magento_functions.sh location with if https://github.com/continuouspipe/dockerfiles/pull/297/files#diff-4a68f150fec1a766024815e12719d4f0R373 is merged in.

magento2-nginx/fpm-7.0/etc/confd/templates/magento/echo_reports_cron.tmpl
```diff
@@ -0,0 +1,5 @@
+# don't send any mail
+MAILTO=""
+SUPERVISORCTL="supervisorctl -c /etc/supervisor/supervisord.conf -u supervisor -p supervisor"
+
+* * * * * root $SUPERVISORCTL start echo_magento_reports
```
sroze: Shouldn't we be able to disable this?

magento2-nginx/fpm-7.0/etc/confd/templates/magento/cron.tmpl
```diff
@@ -2,4 +2,6 @@
 MAILTO=""
 SUPERVISORCTL="supervisorctl -c /etc/supervisor/supervisord.conf -u supervisor -p supervisor"
 
+{{ if eq "true" (getenv "RUN_MAGENTO_CRON") }}
```
sroze: Not sure but is this environment variable documented?

magento2-nginx/fpm-7.0/etc/confd/templates/magento/echo_reports_cron.tmpl
kierenevans: Added a variable for it 👍

magento2-nginx/fpm-7.0/etc/confd/templates/magento/cron.tmpl
kierenevans: It is now! Good spot

PR #305, opened 2017-06-13
Allow choice of image version for Magento2

kierenevans:
Depends on #297 to provide the functions used here.

Depending on the $IMAGE_VERSION variable, use the new setup approach where a database is created in the build stage, magento does lots of things, then the database is removed.

To support both versions of setup, hem is still installed in this base image.


magento2-nginx/fpm-7.0/README.md
```diff
@@ -63,6 +64,7 @@ The following variables are supported
 
 Variable | Description | Expected values | Default
 --- | --- | --- | ----
+IMAGE_VERSION | The docker image version to use. Version 1 uses the install_magento*.sh scripts which can be hard to customise. Version 2 uses magento_functions.sh and does a temporary database installation during the "build" phase. | 1/2 | 1
```
sroze: Shouldn't we default to 2?

kierenevans: Not if we want to keep backward compatibility with the currently deployed images.

PR #312, opened 2017-06-23
Fix mailcatcher ports

kierenevans:
Fix where if a service is called "http" or "smtp", docker/kubernetes provides the IP address of the linked service as "tcp://<ip>:<port>"

### DIFF
--- a/mailcatcher/usr/local/bin/mailcatcher.sh
+++ b/mailcatcher/usr/local/bin/mailcatcher.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
-export HTTP_PORT=${HTTP_PORT:-1080}
-export SMTP_PORT=${SMTP_PORT:-1025}
+HTTP_PORT=${HTTP_PORT:-1080}
+SMTP_PORT=${SMTP_PORT:-1025}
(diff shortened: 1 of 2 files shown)


mailcatcher/usr/local/bin/mailcatcher.sh
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
-export HTTP_PORT=${HTTP_PORT:-1080}
-export SMTP_PORT=${SMTP_PORT:-1025}
+HTTP_PORT=${HTTP_PORT:-1080}
+SMTP_PORT=${SMTP_PORT:-1025}
+if [[ "$HTTP_PORT" =~ "tcp://" ]]; then
```
sroze: Can you add a comment here explaining the reason (that you put on the PR) ?

andytson-inviqa: sounds like a `you're using it wrong` situation

andytson-inviqa: instead perhaps should error if not a number between 1 and 65535

PR #313, opened 2017-06-26
Fix template path for the protect admin feature of Magento2

kierenevans:
Template path didn't exist and stopped confd from starting!


sroze:
😅

PR #316, opened 2017-06-27
Remove load from redis slaves for magento2

kierenevans:
We were encountering pages that were blank due to corrupted layout cache.
We tracked it down to this setting - layout config was loading from the slaves, being edited, then saved to the master, leading to replication lag having an effect.

### DIFF
--- a/magento2-nginx/fpm-7.0/etc/confd/templates/magento/env.php.tmpl
+++ b/magento2-nginx/fpm-7.0/etc/confd/templates/magento/env.php.tmpl
@@ -18,7 +18,6 @@ return array (
                                     'timeout' => '0.5',
                                     'sentinel_master' => '{{ getenv "REDIS_SENTINEL_MASTER" }}',
                                     'sentinel_master_verify' => '1',
-                                    'load_from_slaves' => '2',
 {{ else }}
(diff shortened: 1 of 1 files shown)


sroze:
Does this mean M2 won't use the slaves anymore?

kierenevans:
Yes, after this PR, the redis slaves will be just for failover rather than reading.

We were encountering 0 byte responses and missing header/footer responses with this enabled that after a lot of debugging we found to be a corrupt structured layout cache.

andytson-inviqa:
doe M2 have a concept of a read connection and write connection, like it does mysql? Where it uses the write connection for writes that should be read in the same request?

kierenevans:
Cluster support for M2 is very new (but will be a feature in 2.2.x). It doesn't appear so, looking at the example config for cluster support: https://github.com/colinmollenhour/Cm_Cache_Backend_Redis#high-availability-and-load-balancing-support

I'll raise a bug on Cm_Cache_Backend_Redis in a little bit regarding the bad responses.

sroze:
@kierenevans can you paste the bug ticket on `Cm_Cache_Backend_Redis` here?

PR #317, opened 2017-06-27
Add force_standalone option for magento2's redis configuration.

kierenevans:
Work around issues with phpredis and magento.


magento2-nginx/fpm-7.0/README.md
```diff
@@ -108,6 +108,7 @@ REDIS_SENTINEL_SERVICE_PORT | The port of the redis sentinel service. Used whils
 MAGENTO_REDIS_CACHE_DATABASE | Redis database number to store block cache | database number | 0
 MAGENTO_REDIS_FULL_PAGE_CACHE_DATABASE | Redis database number to store full page cache | database number | 1
 MAGENTO_REDIS_SESSION_DATABASE | Redis database number to store sessions | database number | 2
+MAGENTO_REDIS_FORCE_STANDALONE | Should the extension phpredis be used, or should Credis be forced to use it's own standalone implementation of the redis protocol. Set to 1 (use the standalone implementation) as it is more able to handle connection problems than phpredis | 0/1 | 1
```
sroze: Shouldn't we use `true` and `false` ?

kierenevans: Taking a look at https://github.com/colinmollenhour/Cm_Cache_Backend_Redis/blob/1.10.2/Cm/Cache/Backend/Redis.php#L129 it checks for empty values meaning false, so we could use true/false, interpret this value and convert it in the template 👍

PR #323, opened 2017-06-29
Production mode static file caching for magento2.

kierenevans:
Uncomment the expires max line for production mode caching of static assets in varnish


magento2-nginx/fpm-7.0/etc/confd/templates/nginx/site_base.conf.tmpl
```diff
@@ -76,8 +76,9 @@ location /pub/ {
 }
 
 location /static/ {
-    # Uncomment the following line in production mode
-    # expires max;
+    {{ if eq "production" (getenv "MAGENTO_MODE") }}
+    expires max;
+    {{ end }}
```
andytson-inviqa: should it be clearer by reversing it to a not equal to dev? Otherwise it sounds like it only is applied to the production environment

kierenevans: Great idea, thank you!

kierenevans:
After offline discussion, we are going to link this to a separate variable, as cache-busting is opt-in and this setting right now will cause issues.

magento2-nginx/fpm-7.0/usr/local/share/env/30-framework
```diff
@@ -128,3 +128,7 @@ export MAGENTO_RUN_BUILD
 export MAGENTO_ADMIN_USERNAME=${MAGENTO_ADMIN_USERNAME:-}
 export MAGENTO_ADMIN_PASSWORD=${MAGENTO_ADMIN_PASSWORD:-}
 export MAGENTO_ADMIN_EMAIL=${MAGENTO_ADMIN_EMAIL:-admin@example.com}
+
+MAGENTO_CACHE_STATIC_ASSETS=${MAGENTO_CACHE_STATIC_ASSETS:-false}
+MAGENTO_CACHE_STATIC_ASSETS="$(convert_to_boolean_string "$MAGENTO_CACHE_STATIC_ASSETS")"
```
andytson-inviqa: btw, though I'm not going to hold up this PR about this, but I've been converting these convert_to_boolean_string defaults to be less verbose e.g.:
```
MAGENTO_CACHE_STATIC_ASSETS="$(convert_to_boolean_string "${MAGENTO_CACHE_STATIC_ASSETS:-false")"
```

PR #324, opened 2017-07-03
Add SSL server cipher/protocol/session cache/oscp stapling configuration

andytson-inviqa:


andytson-inviqa:
I've tested and fixed some issues in this, seems ok now

andytson-inviqa:
actually, a Nginx error

andytson-inviqa:
fixed

sroze:
I've no idea what this is about. So @kierenevans I let you have a look 😄

PR #326, opened 2017-07-10
Add instructions for using the SSH functionality.

kierenevans:


ubuntu/16.04/README.md
```diff
@@ -188,6 +188,16 @@ alter any file in the codebase.
...
 
+### SSH Key
+
+The docker images support configuring an SSH key for the "build" user, which is the user that should be running installation tools.
+
+Assuming you have generated a passwordless SSH keypair on your machine (ideally this would be a keypair especially for use in the docker setup, rather than your own), you can do the following to provide the keypair to this docker image:
+
+* `BUILD_USER_SSH_PRIVATE_KEY` is the output of `base64 <PRIVATE_KEY_FILEPATH>`
+* `BUILD_USER_SSH_PUBLIC_KEY` is the output of `base64 <PUBLIC_KEY_FILEPATH>`
+* `BUILD_USER_SSH_KNOWN_HOSTS` is the output of `ssh-keyscan -H -t rsa,ecdsa <HOSTNAME_TO_CONNECT_TO> | base64`, where HOSTNAME_TO_CONNECT_TO is for example github.com or bitbucket.org or any other server.
```
SeerUK: It may also be worth putting a note about potentially needing to include the port with `-p`, if the SSH server you're connecting to is not using port 22.

andytson-inviqa: I don't think it's worth it going through ssh options for edge cases, there's other ways of finding it out

PR #329, opened 2017-07-23
Add the Symfony Flex base image

sroze:


sroze:
@kierenevans @andytson-inviqa WDYT ? Should we actually merge with the Symfony image, not creating a flex one?

symfony-flex/usr/local/share/symfony/symfony_flex_functions.sh
```diff
@@ -0,0 +1,42 @@
...
+    do_symfony_console doctrine:migrations:migrate
+  fi
+
+  HAS_DOCTRINE_ORM=$(has_package doctrine/orm)
+  if [ "$HAS_DOCTRINE_ORM" = "true" ]; then
+    do_symfony_console doctrine:schema:update --force
+  fi 
+}
+
+has_package() {
```
andytson-inviqa: this could be put in the php images for general use

continuouspipe-starter:
> as discussed, we think that there's no need for a separate flex image

@andytson-inviqa including the npx encore things? Should we put them in the Symfony base images?

sroze:
> as discussed, we think that there's no need for a separate flex image

@andytson-inviqa including the npx encore things? Should we put them in the Symfony base images?

sroze:
Oy oy oy! So.. what about this question? https://github.com/continuouspipe/dockerfiles/pull/329#issuecomment-330056736 :)

sroze:
Closed in favour of #457

PR #331, opened 2017-07-23
Add support for postgres assets database import

andytson-inviqa:


andytson-inviqa:
I'm not really sure about what to do with installing the postgres cli/php extension. Installing both the mysql and postgres tools in the same image is wasteful, yet having separate images is too messy

andytson-inviqa:
I suppose this is a good reason to separate the assets code from the PHP image

sroze:
> Installing both the mysql and postgres tools in the same image is wasteful

I wouldn't say so, it's not that bad, and very convenient for users :)

andytson-inviqa:
yep, convenient for users, but not for servers :) I'll make it install both (and that'll fix the docker lint checks)

andytson-inviqa:
I've tested and it works against the project I was using this for

PR #332, opened 2017-07-24
Add new functions for faster permissions setting

kierenevans:
* Add common methods to set permissions so we can extend it in the future to include setfacl if possible.
* Use find to speed up permissions setting by finding files and directories not owned by the right user/group or not the right permissions - this helps a lot when using volumes!

### DIFF
--- a/ubuntu/16.04/usr/local/share/bootstrap/common_functions.sh
+++ b/ubuntu/16.04/usr/local/share/bootstrap/common_functions.sh
@@ -235,3 +235,38 @@ function test_remote_ports() {
     timeout 1 bash -c "cat < /dev/null > /dev/tcp/${SERVICE_PARAMS[0]}/${SERVICE_PARAMS[1]}" 2>/dev/null || return 1
   done
 }
+
+function do_ownership() {
+  local OWNERSHIP_PATH=($1)
+  local USER="$2"
+  local GROUP="$3"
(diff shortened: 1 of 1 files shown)


sroze:
That's a very good idea! What's blocking this PR?

ubuntu/16.04/usr/local/share/bootstrap/common_functions.sh
```diff
@@ -235,3 +235,38 @@ function test_remote_ports() {
...
+
+function do_ownership() {
+  local OWNERSHIP_PATH=($1)
+  local USER="$2"
+  local GROUP="$3"
+  local ALLOW_GROUP_WRITE="${4:-true}"
+  if [ "$IS_CHOWN_FORBIDDEN" != 'true' ]; then
+    find "${OWNERSHIP_PATH[@]}" \( ! -user "${USER}" -or ! -group "${GROUP}" \) -exec chown "${USER}:${GROUP}" {} +
+  fi
+  do_read_write_permissions "${OWNERSHIP_PATH[@]}" "${ALLOW_GROUP_WRITE}"
```
kierenevans: This doesn't pass through the paths correctly to `do_read_write_permissions`, only the first one.

sroze:
@kierenevans do you think you'd have some time at some point to go forward with this issue? Waiting a long time at every container start on the `chown -R build:build /app/` command is quite a disruption on projects.

PR #333, opened 2017-07-25
Update README for PHP 7.1 usage

richdynamix:
Although the README has been updated to reflect the changes made, the shared images would still need to be created and uploaded. 

I am not 100% sure if this is the most optimal way of doing it but the built image is only 9MB larger than the PHP 7.0 version.

### DIFF
--- a/php-nginx/Dockerfile
+++ b/php-nginx/Dockerfile
@@ -4,12 +4,18 @@ MAINTAINER Kieren Evans <kieren.evans+cp-dockerfiles@inviqa.com>
 
 # Install PHP packages, including the Tideways extension
 ARG PHP_VERSION
-ENV PHP_VERSION=${PHP_VERSION:-7.0}
+ENV PHP_VERSION=${PHP_VERSION:-7.1}
 RUN echo 'deb http://s3-eu-west-1.amazonaws.com/qafoo-profiler/packages debian main' > /etc/apt/sources.list.d/tideways.list \
  && curl -sS https://s3-eu-west-1.amazonaws.com/qafoo-profiler/packages/EEB5E8F4.gpg | apt-key add - \
  && if [ "$PHP_VERSION" != "7.0" ]; then \
    echo 'deb http://ppa.launchpad.net/ondrej/php/ubuntu xenial main' > /etc/apt/sources.list.d/php-ppa.list; \
    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E5267A6C; \
+ fi \
+ && if [ "$PHP_VERSION" == "7.1" ]; then \
+    export LANG=C.UTF-8 \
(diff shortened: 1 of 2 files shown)


php-nginx/Dockerfile
```diff
@@ -4,13 +4,19 @@ MAINTAINER Kieren Evans <kieren.evans+cp-dockerfiles@inviqa.com>
 
 # Install PHP packages, including the Tideways extension
 ARG PHP_VERSION
-ENV PHP_VERSION=${PHP_VERSION:-7.0}
+ENV PHP_VERSION=${PHP_VERSION:-7.1}
```
andytson: no need for this, please revert as 7.0 is LTS, and this doesn't even get the default used anyway

php-nginx/Dockerfile
```diff
@@ -4,13 +4,19 @@ MAINTAINER Kieren Evans <kieren.evans+cp-dockerfiles@inviqa.com>
...
  && if [ "$PHP_VERSION" != "7.0" ]; then \
    echo 'deb http://ppa.launchpad.net/ondrej/php/ubuntu xenial main' > /etc/apt/sources.list.d/php-ppa.list; \
    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E5267A6C; \
  fi \
+ && if [ "$PHP_VERSION" == "7.1" ]; then \
+    export LANG=C.UTF-8 \
+    && apt-get update \
+    && apt-get install -y software-properties-common python-software-properties \
+    && add-apt-repository ppa:ondrej/php; \
+ fi \
```
andytson: no need for this, as the other ppa has php 7.1 as well

andytson:
there was already PHP 7.1 support

php-nginx/Dockerfile
richdynamix: I tried to set this with the other one but it was failing to find the packages. Possibly not setting it correctly

andytson: it's fine, it works and has an image built already https://github.com/continuouspipe/dockerfiles/blob/master/docker-compose.yml#L277-L284

richdynamix: Ah ignore my PR then 😝 Should I keep in place the README updates?

kierenevans:
Hey @richdynamix, thank you!
We forgot to update the docs when we started building 7.1, sorry!
If you use `quay.io/continuouspipe/php7.1-nginx:stable` it should have you covered already.
Thank you for updating the docs and nice to see a familiar username :)

richdynamix:
No problem at all. I should probably just make a note to check the `docker-compose.yml` first for any updates :) And I think you might hear from me more. I am working on getting the new company setup with CP. A lot of housekeeping to do first though :D

PR #336, opened 2017-07-31
Spryker: Two vhosts - one for zed and one for yves

kierenevans:
Here we create a copy of the "default" vhost, to allow us to inherit the goodness from the php-nginx parent image.
We then allow this new vhost to use slightly different variables to the default version.

`site_base.conf.tmpl` is customised from php-nginx to remove the "root" statement.

README.md to follow later as it appears to be missing currently(!)


andytson-inviqa:
if it wasn't for GCE volume issues I'd have just suggested separate containers for each. Production should ideally be able to scale each separately

kierenevans:
In the config I've got going, remote dev will be one container which handles both and for other environments there will be a separate set of containers for yves and zed.

andytson-inviqa:
@sroze I want to see if the single vhost way would work before merging this

sroze:
@andytson-inviqa sure; are you testing it?

spryker/nginx/usr/local/share/env/25-framework
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+export WEB_INCLUDES=${WEB_INCLUDES:-default-* yves-*}
+export ZED_WEB_INCLUDES=${ZED_WEB_INCLUDES:-default-01-site_ssl default-05-custom_scheme_flags default-05-site_healthcheck default-06-site_redirect_to_https default-10-base default-20-rewriteapp default-30-phpfpm default-50-disable_further_php zed-*}
```
andytson-inviqa: why not `default-* zed-*`? that would be more stable

andytson-inviqa: i.e. move whatever isn't to yves

andytson-inviqa:
I'd like for a way to turn this off, as in a split up environment, you wouldn't want Zed to be routable from the Yves container and might have security implications

spryker/nginx/usr/local/share/env/25-framework
kierenevans: Just based on the previous discussion regarding choosing the imports. I can update to be `default-* zed-*`

andytson-inviqa:
Apache image appears to work correctly and APP_SERVICES removes undeclared ones correctly, so I think if somone can just review the last 3 commits we can merge this

PR #337, opened 2017-07-31
NGINX: Allow customising the server name and additional includes

kierenevans:
This allows for some pre-templating copying/sed in subsequent images without
having to duplicate all of the content, as used in #336


php-nginx/etc/confd/templates/nginx/site.conf.tmpl
```diff
@@ -21,6 +21,9 @@ server {
     set $access_log_enabled 1;
 
     include /etc/nginx/sites-available/default-*.conf;
+    {{ if ne "" (getenv "WEB_ADDITIONAL_INCLUDES_NAME") }}
+        include /etc/nginx/sites-available/{{ getenv "WEB_ADDITIONAL_INCLUDES_NAME" }}-*.conf;
+    {{ end }}
```
andytson-inviqa: why can't project's just use the existing `default-*`?

kierenevans: To be able to use default-* and another thing where the other thing varies when multiple vhosts are involved.

php-nginx/etc/confd/templates/nginx/site.conf.tmpl
```diff
@@ -1,5 +1,5 @@
 server {
-    server_name _;
+    server_name {{ getenv "WEB_SERVER_NAME" }};
```
andytson-inviqa: what does this functionality give in a single nginx server context? or is it for if adding multiple?

kierenevans: Only for when there are multiple vhosts.

php-nginx/etc/confd/templates/nginx/site.conf.tmpl
andytson-inviqa: I suppose the question is then what in the default-* isn't applicable to the other vhosts (given things like even ssl configuration is in there. Perhaps some thought as to which files from the base image should be always used (or overridable per server)

nginx/README.md
```diff
@@ -79,6 +80,8 @@ WEB_SSL_SESSION_CACHE | Sets the types and sizes of caches that store session pa
 WEB_SSL_SESSION_TIMEOUT | Specifies a time during which a client may reuse the session parameters. | time | 5m
 WEB_SSL_TRUSTED_CERTIFICATES | The trusted certificates to use for OSCP stapling verification and/or SSL client certificate authentication | absolute filename | 
 NGINX_LOG_FORMAT_NAME | Which log format to use for the access log. Two are currently available: combined (provided by NGINX) or combined_with_x_forwarded_for which logs the whole X-Forwarded-For header. | string | combined
+WEB_INCLUDES | A space separated list of files in /etc/nginx/sites-enabled/ to include. ".conf" will be appended automatically. Globs are accepted. | space separated list of partial file names | default-*
```
andytson-inviqa: I'm going to merge, but I would like to somehow show that this is rarely otherwise needed.

PR #338, opened 2017-08-03
Spryker latest demoshop improvements - hostname changes

kierenevans:
Building upon #336 and #337, a raft of changes were required to get the spryker image working with a newer demoshop version.


spryker/nginx/usr/local/share/container/baseimage-45.sh
```diff
@@ -7,12 +7,15 @@ do_build_permissions() {
...
 }
 
-alias_function do_composer do_spryker_composer_inner
-do_composer() {
-  do_spryker_composer_inner
-  do_spryker_app_permissions
-  do_build_assets
+alias_function do_build do_spryker_nginx_build_inner
+do_build() {
+  export DEVELOPMENT_MODE="${BUILD_DEVELOPMENT_MODE}"
```
andytson-inviqa: This would be replaced by the upstream image's RUN_BUILD

andytson-inviqa: actually, nope, just a DEVELOPMENT_MODE build arg, no need for a BUILD_ one

spryker/nginx/usr/local/share/env/25-framework
```diff
@@ -1,18 +1,46 @@
...
+
+export ELASTICSEARCH_HOST=${ELASTICSEARCH_HOST:-elasticsearch}
+export ELASTICSEARCH_HOST_PORT=${ELASTICSEARCH_HOST_PORT:-9200}
+export REDIS_HOST=${REDIS_HOST:-redis}
+export REDIS_HOST_PORT=${REDIS_HOST_PORT:-6379}
+export MAILCATCHER_HOST=${MAILCATCHER_HOST:-mailcatcher}
+export MAILCATCHER_HOST_PORT=${MAILCATCHER_HOST_PORT:-1080}
+
+export YVES_HOST=${YVES_HOST:-yves}
+export YVES_HOST_PROTOCOL=${YVES_HOST_PROTOCOL:-https://}
```
sroze: Can we remove `://` from here? Technically that's not really part of the protocol.

spryker/nginx/usr/local/share/env/25-framework
```diff
@@ -1,18 +1,46 @@
...
+export ZED_HOST=${ZED_HOST:-zed}
+export ZED_HOST_PROTOCOL=${ZED_HOST_PROTOCOL:-https://}
+
+export WEB_SERVER_NAME=${WEB_SERVER_NAME:-"yves $YVES_HOST"}
```
sroze: So by default it's called "yves yves" ?

spryker/nginx/usr/local/share/env/25-framework
andytson-inviqa: also can use php parseurl to split it up so the user doesn't have to define so many variables

spryker/nginx/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -60,33 +66,40 @@ do_generate_files() {
...
+    SPRYKER_INSTALLED=$?
+  set -e
+
+  if [ "$SPRYKER_INSTALLED" -ne 1 ]; then
+    as_code_owner "vendor/bin/console setup:install"
+    do_import_demodata
+    do_propel_install
+    do_product_label_relations_update
+    do_run_collectors
+    do_setup_search
```
hgajjar: do_setup_search should be executed before run_collectors

spryker/nginx/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -60,33 +66,40 @@ do_generate_files() {
...
+  # check if spryker is installed
+  set +e
+    psql -U "$DATABASE_USER" -h "$DATABASE_HOST" -c "SELECT EXISTS (SELECT * FROM   information_schema.tables WHERE table_catalog = '$DATABASE_NAME' AND table_name = 'spy_locale');" | grep -q f
+    SPRYKER_INSTALLED=$?
+  set -e
+
+  if [ "$SPRYKER_INSTALLED" -ne 1 ]; then
+    as_code_owner "vendor/bin/console setup:install"
+    do_import_demodata
+    do_propel_install
```
hgajjar: I think propel install is not required here, and even if its required then it should come before demodata install

spryker/nginx/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -60,33 +66,40 @@ do_generate_files() {
...
+    do_run_collectors
+    do_setup_search
+  fi
+}
+
+do_spryker_migrate() {
+  do_propel_install
+
+  if [ "$APPLICATION_ENV" != "production" ]; then
+    do_run_collectors
```
hgajjar: this is only required if we don't run crons by default for non-production envs, as the collectors are already being run by cron, Also if cron is already running collectors then this will conflict with that

spryker/nginx/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -95,10 +108,37 @@ do_run_collectors() {
   as_code_owner "vendor/bin/console collector:storage:export"
 }
 
+do_propel_install() {
+  as_code_owner "vendor/bin/console propel:install"
```
hgajjar: add "-o" flag to this

spryker/nginx/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -95,10 +108,37 @@ do_run_collectors() {
...
+do_product_label_relations_update() {
+  as_code_owner "vendor/bin/console product-label:relations:update"
 }
 
 do_setup_search() {
   as_code_owner "vendor/bin/console setup:search"
 }
+
+do_run_tests() {
+  as_code_owner "vendor/bin/codecept run --skip Acceptance"
```
hgajjar: may be this shouldn't skip acceptance test in default image

spryker/nginx/etc/confd/templates/nginx/site_internal_requests.conf.tmpl
```diff
@@ -0,0 +1,3 @@
+if ($http_x_internal_request ~ 1) {
```
andytson-inviqa: how do we ensure this isn't set by anything external?

spryker/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -101,10 +114,37 @@ do_run_collectors() {
...
+do_spryker_hosts() {
+    set +e
+    grep -q zed /etc/hosts
+    HOSTS_UPDATED=$?
+    set -e
+    if [ "$HOSTS_UPDATED" != "0" ]; then
+        # zed is required for internal endpoint and others are required for Acceptance tests to not loop over internet in order to resolve those endpoints
+        echo "127.0.0.1 zed $ZED_HOST $YVES_HOST" >> /etc/hosts
+    fi
+}
```
andytson-inviqa: this may need some adjustment based on APP_SERVICES

spryker/nginx/usr/local/share/env/25-framework
andytson-inviqa: I've found this is how spryker expects it's config input. Doesn't mean we can't simplify it though

spryker/nginx/usr/local/share/env/25-framework
andytson-inviqa: yes, not really an issue though

spryker/nginx/usr/local/share/spryker/spryker_functions.sh
andytson-inviqa: yep, I didn't need to run it, so I'll remove it

andytson-inviqa: I expect setup:install does it

spryker/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -38,73 +46,126 @@ do_spryker_app_permissions() {
...
+    do_spryker_product_label_relations_update
+    do_spryker_run_collectors
+    do_spryker_setup_search
+  fi
+}
+
+do_spryker_migrate() {
+  do_spryker_propel_install
+
+  if [ "$APPLICATION_ENV" != "production" ]; then
```
andytson-inviqa: A danger here in that if production is only used in production then this process is only ever skipped in production and never tested skipped before production

spryker/usr/local/share/env/30-framework
```diff
@@ -31,9 +31,15 @@ export YVES_HOST_PROTOCOL=${YVES_HOST_PROTOCOL:-https://}
 export ZED_HOST=${ZED_HOST:-zed}
 export ZED_HOST_PROTOCOL=${ZED_HOST_PROTOCOL:-https://}
 
+export YVES_WEB_SERVER_NAME=${YVES_WEB_SERVER_NAME:-"yves $YVES_HOST"}
+export ZED_WEB_SERVER_NAME=${ZED_WEB_SERVER_NAME:-"zed $ZED_HOST"}
```
kierenevans: Previously with the vhost split, there was nothing listening to all domain names I think.

andytson-inviqa: I'm not sure what you mean. If YVES_HOST points at 127.0.0.1 and is used in the server name, it would be the same as `yves` pointing to 127.0.0.1 and used in the server name

andytson-inviqa: so, wouldn't it be possible just to use $YVES_HOST instead of `yves` for all configurations?

sroze:
That seams very reasonable. As you guys worked together on the last commits, I guess you agree with each other. Should we merge?

spryker/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -157,6 +157,17 @@ do_spryker_run_tests() {
...
 }
 
+do_spryker_hosts() {
+    set +e
+    grep -q zed /etc/hosts
+    HOSTS_UPDATED=$?
+    set -e
+    if [ "$HOSTS_UPDATED" != "0" ]; then
+        # zed is required for internal endpoint and others are required for Acceptance tests to not loop over internet in order to resolve those endpoints
+        echo "127.0.0.1 zed $ZED_HOST $YVES_HOST" >> /etc/hosts
```
sroze: This should not be mandatory. Some platforms refuse to write in this file.

andytson-inviqa: so instead we could just use the k8s dns for CP and use docker-compose extra_hosts/link aliases

andytson-inviqa: hmm, actually, can't do the k8s dns way for when on the same host, as it's set for the multiple vhosts, and k8s dns is one dns address per service I guess

sroze:
Should we close this PR?

spryker/nginx/etc/confd/templates/nginx/site_internal_requests.conf.tmpl
kierenevans: Seems that setting `$config[ZedRequestConfig::IGNORE_SSL_CERT_SELF_SIGNING] = true;` allows us to talk to https.

PR #339, opened 2017-08-07
Backport Symfony app endpoint code to php images

andytson-inviqa:
Since PHP images are effectively doing the same thing, and other images aren't really related to Symfony but extending it just for this functionality, move it to the PHP images so it's more accessible.

Note there is a non-BC change here in the PHP image, in that the image no longer has the app endpoint enabled by default. This is because most


andytson-inviqa:
tested and fixed a few issues, but also I:
* renamed APP_ENDPOINT_REDIRECT to APP_ENDPOINT_REWRITE (rebased into the original commit)
* changed the strict mode to hide the PHP files so as if they weren't there

PR #340, opened 2017-08-08
Make spryker image from php images and add apache

andytson-inviqa:
Spryker doesn't use symfony, just symfony components, so the symfony image is mostly redundant.

Also no reason not to have a Apache image


andytson-inviqa:
I've added https://quay.io/repository/continuouspipe/spryker-php7.1-apache?tab=info with the robot user

PR #341, opened 2017-08-08
Fix solr6 infinite loop.

kierenevans:
Solr's docker entry point brings in every file in /docker-entrypoint-initdb.d/*,
so calling the docker-entrypoint.sh script manually led to an infinite loop.


sroze:
Could you instead return if the `IS_INNER_REQUEST` variable exists? Instead of nesting it? (with a comment explaining the use case)

kierenevans:
Updated!

PR #343, opened 2017-08-08
Add EZ PHP 7.1 image

kierenevans:
EZ Platform uses Symfony framework (not just Symfony components), so will benefit from using the Symfony images as the base images.


ez/6.x/Dockerfile-php7.1
```diff
@@ -0,0 +1,6 @@
+FROM quay.io/continuouspipe/symfony-php7.1-apache:latest
```
andytson-inviqa: how much of Symfony does it use? If not supporting the symfony bundle system I think we should do as I did with Spryker. All the symfony_functions.sh assume there's the symfony console tool

ez/6.x/Dockerfile.tmpl
```diff
@@ -0,0 +1,6 @@
+FROM quay.io/continuouspipe/symfony-php{{PHP_VERSION}}-apache:latest
```
andytson-inviqa: how much of Symfony does it use? If not supporting the symfony bundle system I think we should do as I did with Spryker. All the symfony_functions.sh assume there's the symfony console tool

andytson-inviqa: looks like it does use it properly, so can ignore this

continuous-pipe.yml
```diff
@@ -110,6 +116,9 @@ tasks:
         ez:
```
andytson-inviqa: would be good to rename this one for consistency

ez/6.x/Dockerfile.tmpl
kierenevans: Yep, very much a framework built from the full symfony stack.

continuous-pipe.yml
kierenevans: Updated

PR #346, opened 2017-08-09
Run assets apply for the database during "setup"

kierenevans:
"setup" is called from a separate container usually to avoid web startup times being long.


andytson-inviqa:
should we just put this in the php image instead?

kierenevans:
Sounds good!

PR #348, opened 2017-08-09
Add Sqlite to PHP base images

sroze:


andytson-inviqa:
why? I'm not trying to block it, but I asked why you temporarily added it elsewhere when I didn't run into issues

sroze:
In the context of Symfony, what some user do is to run tests with the sqlite database backend for Doctrine, so they don't need to deploy the actual database. I think it's a nice idea and this wasn't supported because of the missing driver

andytson-inviqa:
If they use migrations for data or fixtures that are using mysql features, that'll break, but ok as it it wont hurt if not

PR #353, opened 2017-08-10
Run assets apply during the setup phase, but at PHP image level.

kierenevans:
Similar to #346 but in a higher base image.


andytson-inviqa:
better encapsulation

PR #355, opened 2017-08-10
Enable mod_remoteip for apache

kierenevans:
Allows apache to remove trusted reverse proxies from X-Forwarded-For like nginx does.


kierenevans:
Getting some wrong results with this still, so will need further looking into.

kierenevans:
The wrong results are the logs - we need to change the log format to use %a instead - for another PR.

PR #356, opened 2017-08-11
Spryker latest demoshop improvements - excluding hostname changes

andytson-inviqa:


andytson-inviqa:
I created this so this stuff to push the work forward whilst bits are discussed and possibly changed

PR #357, opened 2017-08-11
ez images are ezplatform

andytson-inviqa:
Too late to change the image names though


sroze:
The files are `ezplatform/6.x/[...]` now... "eZ Platform 6.0" is not a thing either!

PR #362, opened 2017-08-16
Add a has_composer_package function to php images

andytson-inviqa:
Credits to @sroze for creating it in #329


andytson-inviqa:
just seen it's use-case uses "true" but I want to convert to return values for simple `if has_package "package"` bash evaluation

andytson-inviqa:
this doesn't currently check packages-dev, which would be a good idea also

symfony/usr/local/share/symfony/symfony_functions.sh
```diff
@@ -58,12 +98,18 @@ do_database_build() {
 
 do_database_install() {
   do_database_schema_create
-  do_database_migrations_mark_done
+  if uses_symfony_doctrine_mode_migrations; then
+    do_database_migrations_mark_done
```
sroze: I think we should run the migration script instead of running the database schema create. The reason for this is that some application will have migrations that are not 100% in sync with the doctrine generated schema, and that's exactly why they use the migrations instead of schema:update.

sroze:
Great work tho 👍

symfony/usr/local/share/symfony/symfony_functions.sh
andytson-inviqa: ok, fair point. I did this as migrations could take much longer, but we could just rely on the s3 assets functionality to speed them up. Migrations may not even be ORM related, so running them in favour of the schema seems like the right idea when in migrations mode

PR #366, opened 2017-08-18
Magento 2.2 Support

kierenevans:
A few things have changed regarding validation of CLI parameters - database password cannot be empty, so setup:install fails during the docker build.

The `app` directory needs to be owned by the user running composer, to be able to create `bin/magento` later on (otherwise it silently exits).

Magento 2.2 adds a `generated` folder in the root of the codebase, which needs to be writeable by www-data like `var/generation/` used to be.

In gearing up for local development, we need to wait for the database to be available before running the installer or restoring a database.

Currently there is a setup bug with a few modules in 2.2rc2.1, and we have worked around this by running the installer twice, clearing redis cache in between!
This approach doesn't yet work in the build of the docker image (during `do_magento2_build`) but we are working on a solution for that.

### DIFF
--- a/magento2/usr/local/share/magento2/magento_functions.sh
+++ b/magento2/usr/local/share/magento2/magento_functions.sh
@@ -248,11 +248,10 @@ function do_magento_installer_install() {
 
     echo 'Install Magento 2 via the Installer'
     chmod +x bin/magento
-    as_code_owner "bin/magento setup:install --base-url='$PUBLIC_ADDRESS' \
+    local INSTALL_COMMAND="bin/magento setup:install --base-url='$PUBLIC_ADDRESS' \
       --db-host='$DATABASE_HOST' \
       --db-name='$DATABASE_NAME' \
       --db-user='$DATABASE_USER' \
-      --db-password='$DATABASE_PASSWORD' \
       --admin-firstname=Admin \
       --admin-lastname=Demo \
       --admin-user='${MAGENTO_ADMIN_USERNAME:-admin}' \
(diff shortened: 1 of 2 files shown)


magento2/usr/local/share/magento2/magento_functions.sh
```diff
@@ -376,6 +394,10 @@ function do_magento_download_magerun2() {
 
 function do_magento2_templating() {
   mkdir -p /app/app/etc/
+  chown -R "${CODE_OWNER}:${CODE_GROUP}" /app/app/
+  if [ "${APP_USER_LOCAL}" == "true" ]; then
+    chown "${CODE_OWNER}:${CODE_GROUP}" /app/app/etc/env.php
```
andytson-inviqa: I'm confused why this is needed when the chown above is recursive and a parent directory

kierenevans: Might have been due to us overriding the function in `plan.sh`. Have removed.

magento2/usr/local/share/magento2/magento_functions.sh
```diff
@@ -202,6 +207,9 @@ function do_magento_database_create() {
 
 function do_magento_database_install() {
   set +x
+  if [ "${DATABASE_HOST}" != "localhost" ]; then
+    wait_for_remote_ports "30" "${DATABASE_HOST}:${DATABASE_PORT}"
```
andytson-inviqa: this needs https://github.com/continuouspipe/dockerfiles/pull/374 in order to not cause a regression where mysql service name is called database

PR #367, opened 2017-08-21
Merge the shared sources of php-apache and php-nginx images

andytson-inviqa:


kierenevans:
https://github.com/continuouspipe/dockerfiles/blob/5e7559230982e03e058f94344c36a265c732a6b1/magento2/usr/local/share/container/baseimage-30.sh#L13 will be affected which is used to skip some functionality. I'll put in a fix for that function today.

andytson-inviqa:
it's inadvertently dropping off do_assets_all in addition to the do_build_permissions. Maybe override do_build_permissions instead somehow

sroze:
The problem is: this is not backward compatible. Do we know about any project using these `do_nginx*` or `do_apache*` functions?

andytson-inviqa:
it's backwards compatible, check out the baseimage-21.sh files, which add in the do_nginx and do_apache functions

sroze:
Wow, well done, sorry didn't catch it. Let's go for it then 👍

andytson-inviqa:
the only case of non-BC is with the alias_function "inner" function names, which have changed, but they were never a "public api"

sroze:
Let's wait for @kierenevans to confirm he has put a fix for the usage he knows about then.

andytson-inviqa:
he's put it in, I merged his fix earlier :)

PR #368, opened 2017-08-21
Run magerun in magento1 images as app user

andytson-inviqa:
code owner was causing the log directory and files to be created as build user, when www-data needs write access


andytson-inviqa:
> Permissions are very tricky things!
Well, except in docker for mac with osxfs

PR #370, opened 2017-08-22
Spryker Crons

kierenevans:
Run the Spryker crons without Jenkins.


spryker/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -119,6 +119,7 @@ do_spryker_install() {
     do_spryker_import_demodata
     do_spryker_product_label_relations_update
     do_spryker_setup_search
+    do_spryker_app_permissions
     do_spryker_run_collectors
```
andytson-inviqa: wondering if we should remove this line if we are to rely on crons for dev too

kierenevans: Would mean waiting up to a minute to see something on Yves after the cron container is deployed.

andytson-inviqa: ok

spryker/Dockerfile.tmpl
```diff
@@ -18,4 +18,5 @@ RUN curl -sL https://deb.nodesource.com/setup_6.x > /tmp/install-node.sh \
  && rm -rf /var/lib/apt/lists/*
 
 COPY ./{{WEB_SERVER}}/ /
+COPY ./etc/ /etc/
```
andytson-inviqa: ah, build.sh not run

PR #371, opened 2017-08-22
Reduce all Dockerfile COPY's to a single layer

andytson-inviqa:
This excludes the php images, which are done in another PR


sroze:
What the value of that?

andytson-inviqa:
I don't think layers are as bad as they used to, but less layers in COPY would have some performance boost somewhere

sroze:
I definitely love the other PR but I'm afraid this one is too much of an optimisation that reduces a lot the clarity of these dockerfiles. I personally vote for keeping them as they are, with two `COPY`, just for the sake of better user experience.

andytson-inviqa:
I'm happy to drop this if that's an issue, but what clarity is lost?

sroze:
From a user perspective, this is more or less obvious:
```Dockerfile
COPY ./etc/ /etc
COPY ./usr/ /usr
```

This one is much less:
```Dockerfile
COPY ./shared/ /
```

And these Dockerfiles are also a source of inspiration for the people that are using them, therefore I suggest we keep it very simple.

andytson-inviqa:
the latter style is already being used in some images with nginx/apache varients

andytson-inviqa:
but I disagree that / is less clear than /folder

PR #378, opened 2017-08-24
Remove the obsolete nginx/nginx_site.conf.tmpl

andytson-inviqa:
This was stopped being used very early on, as it didn't promote reuse with each project copying the full template to make changes


kierenevans:
Build for PR had passed but not updated status. Merging!

PR #379, opened 2017-08-24
Create a PHP 5.5 Magento1 EOL image for older Magento versions not patched

andytson-inviqa:


continuous-pipe.yml
```diff
@@ -56,6 +56,13 @@ tasks:
             - name: PHP_VERSION
               value: '7.0'
         php56_nginx:
```
kierenevans: This ID would need to change

continuous-pipe.yml
```diff
@@ -55,6 +55,13 @@ tasks:
           environment:
             - name: PHP_VERSION
               value: '7.0'
+        php55_nginx:
+          image: quay.io/continuouspipe/eol-php5.5-nginx
```
kierenevans: Saved some time and I've created this and the other one 👍

continuous-pipe.yml
andytson-inviqa: fixed

continuous-pipe.yml
andytson-inviqa: thanks

PR #380, opened 2017-08-24
Split Magento 2 vhost

kierenevans:
Split big vhost into smaller chunks to allow overriding some bits instead of all.


andytson-inviqa:
should we ensure the position of the locations are preserved for precedence?

kierenevans:
Good idea, will update :+1:

andytson-inviqa:
I don't think it matters in their current location matches, as any order seems ok for them, just concerned if someone overrides and expects the official M2's nginx order to apply

PR #381, opened 2017-08-25
Branch Builds

andytson-inviqa:


continuous-pipe.yml
```diff
@@ -1,323 +1,394 @@
...
+          environment:
+            PHP_VERSION: '5.6'
+            WEB_SERVER: apache
         tideways:
           image: quay.io/continuouspipe/tideways
-          tag: latest
+          tag: ${FROM_TAG}
           reuse: false
 
 filter: code_reference.branch in ["master"]
```
kierenevans: I re-added this until we could update the all the images - shall we give it a test on this branch by removing this filter?

PR #387, opened 2017-09-01
Add target rule cache to Magento 2

kierenevans:
For enterprise edition installation of Magento 2, there are additional features that can have caches enabled.


tkotosz:
looks good to me :) :+1:

PR #392, opened 2017-09-04
Fix build

kierenevans:
* Fix magento1 builds
* Fix magento2 builds
* Avoid long build times for branch builds.


andytson-inviqa:
Can we have the php 5.5 build running on the very last "fast" build so it's parallelism is at it's most efficient?

kierenevans:
38 minutes is alright(ish), and gives us a target to improve upon! Can't filter the 5.5 images from building without having separate tasks, but then we lose the parallelism for when we do want to build them.

kierenevans:
A much more palatable 11 minutes, thanks @andytson-inviqa for the message!

PR #393, opened 2017-09-07
Include all local IPs in M2 varnish purge acl

fratcliffe:


andytson-inviqa:
I don't like this having docker-compose defaults in in the first place, as they are environment specific. They should go in the docker-compose.yml file

andytson-inviqa:
well, there's nothing wrong with the PR changes, just I have issue with the defaults in general, feel free to merge

PR #374, opened 2017-08-22
Canonize service port env vars so K8S doesn't break them with tcp://<…

andytson-inviqa:
…ip>: prepended

Also reduce _HOST_PORT usage keeping it standard _PORT

PR #394, opened 2017-09-08
Add detection and support for Symfony Flex structure

andytson-inviqa:
depends on #396 and #374 to work


andytson-inviqa:
I've tested this against symfony/skeleton and symfony/framework-standard-edition and they appear to be getting the right environment variables

TerjeBr:
This PR has broken our build!

/usr/local/share/40-stack has this line:
`export APP_ENDPOINT=${APP_ENDPOINT:-/index.php}`

Before DEFAULT_APP_ENDPOINT was computed and used to set APP_ENDPOINT to `/app.php`
in /usr/local/share/env/30-framework
This logic with DEFAULT_APP_ENDPOINT has been moved to /usr/local/share/env/45-framework, but that runs after /usr/local/share/40-stack, so the APP_ENDPOINT now always ends up as /index.php.

andytson:
@TerjeBr sorry to hear that. You can use the :stable tag, which doesn't have this in.

The :stable tag is what we recommend using by default, but the :latest has at least identified this as an issue for you ahead of us tagging :latest as :stable

andytson:
we will probably fix by moving the APP_ENDPOINT in /usr/local/share/40-stack later to avoid the issue

PR #395, opened 2017-09-08
Add spryker_console command for access to spryker commands

rgpjones:
This allows a developer to run any Spryker console command on a CP container, for example if an importer needs to be run.

### DIFF
--- a/spryker/usr/local/share/spryker/spryker_functions.sh
+++ b/spryker/usr/local/share/spryker/spryker_functions.sh
@@ -156,3 +156,12 @@ do_spryker_setup_search() {
 do_spryker_run_tests() {
   as_code_owner "vendor/bin/codecept run --skip Acceptance"
 }
+
+do_spryker_console() (
+  set +x
+  if [ "$#" -gt 0 ]; then
+    as_app_user "vendor/bin/console $(printf "%q " "$@")"
+  else
+    as_app_user "vendor/bin/console"
+  fi
+)


spryker/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -156,3 +156,12 @@ do_spryker_setup_search() {
 do_spryker_run_tests() {
   as_code_owner "vendor/bin/codecept run --skip Acceptance"
 }
+
+do_spryker_console() {
```
kierenevans: As `set +x` is in use on line 161, can we change this to be a subshell execution instead, e.g.
```do_spryker_console() (
  # ...
)``` rather than the `{ }`

sroze:
That's a great contribution @rgpjones, thank you. Could you mention this in the README.md file as well?

spryker/usr/local/share/spryker/spryker_functions.sh
sroze: I don't get that

kierenevans: Avoids the need to reset the turning off of tracing with `set -x` later on, by running everything for the function in a subshell ( see https://github.com/continuouspipe/dockerfiles/pull/365 )

rgpjones:
@sroze What README.md file? There isn't one for spryker right now

sroze:
If there is no Spryker README, then.. thank you :D

PR #396, opened 2017-09-08
Give /usr/local/share/env/ files access to baseimage functions

andytson-inviqa:
by changing the order of load


andytson-inviqa:
appears to have a bug caused by it not having cd /app 'd yet:
jq: error: Could not open file composer.lock: No such file or directory
/usr/local/share/env/45-framework: line 3: [: : integer expression expected

andytson-inviqa:
actually, more an issue with the other PR with a chicken and egg scenario

sroze:
@andytson-inviqa would this be resolved by adding a `WORKDIR /app` in the PHP base image?

andytson:
no, it's resolved now

sroze:
Great. Then if @kierenevans is happy, I guess you can merge :)

PR #399, opened 2017-09-11
Hotfix: Fix spryker running console commands before having run composer.

kierenevans:
Potential fix for seeing vendor/bin/console used before composer install was called.

### DIFF
--- a/spryker/usr/local/share/container/baseimage-45.sh
+++ b/spryker/usr/local/share/container/baseimage-45.sh
@@ -4,7 +4,8 @@ source /usr/local/share/spryker/spryker_functions.sh
 alias_function do_build_permissions do_spryker_build_permissions_inner
 do_build_permissions() {
   do_spryker_build_permissions_inner
-  do_spryker_build
+  do_spryker_directory_create
+  do_spryker_config_create
 }
 
 alias_function do_build do_spryker_nginx_build_inner
--- a/spryker/usr/local/share/container/baseimage-45.sh
+++ b/spryker/usr/local/share/container/baseimage-45.sh
@@ -1,13 +1,6 @@
 #!/bin/bash
 source /usr/local/share/spryker/spryker_functions.sh
 
-alias_function do_build_permissions do_spryker_build_permissions_inner
-do_build_permissions() {
-  do_spryker_build_permissions_inner
-  do_spryker_directory_create
-  do_spryker_config_create
-}
-
 alias_function do_build do_spryker_nginx_build_inner
 do_build() {
   do_spryker_nginx_build_inner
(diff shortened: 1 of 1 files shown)


spryker/usr/local/share/container/baseimage-45.sh
```diff
@@ -4,7 +4,8 @@ source /usr/local/share/spryker/spryker_functions.sh
 alias_function do_build_permissions do_spryker_build_permissions_inner
 do_build_permissions() {
   do_spryker_build_permissions_inner
-  do_spryker_build
+  do_spryker_directory_create
+  do_spryker_config_create
```
andytson-inviqa: if do_spryker_build is running these anyway, do we need to do these here? Also what happens if in dev mode if the other parts of the do_spryker_build aren't run on the mount? Do you think it can be just that the call to do_spryker_build is moved to do_development_start?

spryker/usr/local/share/container/baseimage-45.sh
```diff
@@ -4,11 +4,15 @@ source /usr/local/share/spryker/spryker_functions.sh
...
 }
 
 alias_function do_build do_spryker_nginx_build_inner
 do_build() {
+  export DEVELOPMENT_MODE="${BUILD_DEVELOPMENT_MODE}"
+  export APPLICATION_ENV="${BUILD_APPLICATION_ENV}"
```
andytson-inviqa: BUILD_* is obsolete and never needed, projects can just set DEVELOPMENT_MODE and APPLICATION_ENV

spryker/usr/local/share/env/60-framework
```diff
@@ -4,3 +4,8 @@
 export ELASTICSEARCH_HOST_PORT=${ELASTICSEARCH_PORT}
 export REDIS_HOST_PORT=${REDIS_PORT}
 export MAILCATCHER_HOST_PORT=${MAILCATCHER_PORT}
+
+# Spryker build as development
+export BUILD_DEVELOPMENT_MODE=${BUILD_DEVELOPMENT_MODE:-false}
+export BUILD_APPLICATION_ENV=${BUILD_APPLICATION_ENV:-development}
```
andytson-inviqa: we set symfony env to prod by default, shouldn't we also do the same for spryker?

spryker/usr/local/share/container/baseimage-45.sh
sroze: > Do you think it can be just that the call to do_spryker_build is moved to do_development_start?

I guess that could be the solution. @kierenevans @richardmiller do you think it would be a problem for the local development?

sroze:
@andytson-inviqa you merge?

PR #401, opened 2017-09-11
Avoid conflicting with Magento 2.2 workflow

kierenevans:
Magento 2.2 workflow has app/etc/config.php committed to the repository.
Don't overwrite this with a template.

### DIFF
--- a/magento2/usr/local/share/env/30-framework
+++ b/magento2/usr/local/share/env/30-framework
@@ -180,3 +180,5 @@ function detect_enterprise_edition() (
 )
 MAGENTO_ENTERPRISE_EDITION=$(convert_to_boolean_string "${MAGENTO_ENTERPRISE_EDITION:-$(detect_enterprise_edition)}")
 export MAGENTO_ENTERPRISE_EDITION
+
+export MAGENTO_VERSION="${MAGENTO_VERSION:-2.1}"
--- a/magento2/usr/local/share/env/30-framework
+++ b/magento2/usr/local/share/env/30-framework
@@ -169,16 +169,7 @@ export MAGENTO_ENABLE_CONFIG_WEBSERVICE_CACHE
 MAGENTO_ENABLE_TARGET_RULE_CACHE=$(convert_to_boolean_string "${MAGENTO_ENABLE_TARGET_RULE_CACHE:-true}")
 export MAGENTO_ENABLE_TARGET_RULE_CACHE
 
-function detect_enterprise_edition() (
-  set +e
-  source /usr/local/share/php/common_functions.sh
-  if has_composer_package "magento/product-enterprise-edition"; then
-    echo 'true'
-  else
-    echo 'false'
-  fi
-)
-MAGENTO_ENTERPRISE_EDITION=$(convert_to_boolean_string "${MAGENTO_ENTERPRISE_EDITION:-$(detect_enterprise_edition)}")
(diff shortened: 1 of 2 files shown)


magento2/usr/local/share/env/30-framework
```diff
@@ -180,3 +180,5 @@ function detect_enterprise_edition() (
 )
 MAGENTO_ENTERPRISE_EDITION=$(convert_to_boolean_string "${MAGENTO_ENTERPRISE_EDITION:-$(detect_enterprise_edition)}")
 export MAGENTO_ENTERPRISE_EDITION
+
+export MAGENTO_VERSION="${MAGENTO_VERSION:-2.1}"
```
andytson-inviqa: worth using composer_package_version to find this out?

magento2/usr/local/share/magento2/magento_functions.sh
```diff
@@ -386,6 +386,10 @@ function do_magento_download_magerun2() {
 }
 
 function do_magento2_templating() {
+  if [ "$MAGENTO_VERSION" != "2.0" ] && [ "$MAGENTO_VERSION" != "2.1" ]; then
```
andytson-inviqa: there's also composer_package_compare that can be used here if you can always match a package (for CE and EE)

andytson-inviqa: (or could deconstruct the comparison part)

fratcliffe:
Feedback addressed, we are now auto detecting Magento version :)

magento2/usr/local/share/env/30-framework
```diff
@@ -169,16 +169,7 @@ export MAGENTO_ENABLE_CONFIG_WEBSERVICE_CACHE
...
-  else
-    echo 'false'
-  fi
-)
-MAGENTO_ENTERPRISE_EDITION=$(convert_to_boolean_string "${MAGENTO_ENTERPRISE_EDITION:-$(detect_enterprise_edition)}")
+MAGENTO_ENTERPRISE_EDITION=$(convert_to_boolean_string "${MAGENTO_ENTERPRISE_EDITION:-$(run_return_boolean has_composer_package "magento/product-enterprise-edition")}")
 export MAGENTO_ENTERPRISE_EDITION
 
-export MAGENTO_VERSION="${MAGENTO_VERSION:-2.1}"
+export MAGENTO_VERSION="${MAGENTO_VERSION:-$(composer_package_version magento/product-community-edition | cut -d. -f1,2)}"
```
andytson: composer_package_version will error if it wont match the package. I don't know if we want that behaviour or not for this

magento2/usr/local/share/magento2/magento_functions.sh
```diff
@@ -408,10 +408,15 @@ function do_magento_download_magerun2() {
   chmod +x /app/bin/n98-magerun2.phar
 }
 
-function do_magento2_templating() {
-  if [ "$MAGENTO_VERSION" != "2.0" ] && [ "$MAGENTO_VERSION" != "2.1" ]; then
+function remove_config_template() (
+  set +e
```
andytson: you don't need to `set +e` or sub-shell. the `if` condition keeps the script running

andytson: for instance, how we did `[ "$MAGENTO_VERSION" != "2.0" ]`, this would return an exit status 0/1 based on if it matches. Its the `if` that catches it instead of the -e flag causing it to stop further execution and show the error

PR #402, opened 2017-09-11
Call composer config upon Magento 2 startup.

kierenevans:
In order to support development when composer install has happened in another container, set up ~build/.composer/auth.json on boot.

### DIFF
--- a/magento2/usr/local/share/container/baseimage-30.sh
+++ b/magento2/usr/local/share/container/baseimage-30.sh
@@ -9,6 +9,12 @@ if [ "$IMAGE_VERSION" -ge 2 ]; then
     PRODUCTION_ENVIRONMENT="$BUILD_PRODUCTION_ENVIRONMENT" MAGENTO_MODE="$BUILD_MAGENTO_MODE" DEVELOPMENT_MODE="$BUILD_DEVELOPMENT_MODE" do_magento2_build
   }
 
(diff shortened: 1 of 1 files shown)


magento2/usr/local/share/container/baseimage-30.sh
```diff
@@ -9,6 +9,12 @@ if [ "$IMAGE_VERSION" -ge 2 ]; then
     PRODUCTION_ENVIRONMENT="$BUILD_PRODUCTION_ENVIRONMENT" MAGENTO_MODE="$BUILD_MAGENTO_MODE" DEVELOPMENT_MODE="$BUILD_DEVELOPMENT_MODE" do_magento2_build
   }
 
+  alias_function do_start do_magento2_start_inner
+  do_start() {
+    do_magento2_start_inner
+    do_composer_config
```
andytson-inviqa: How does M2 use this in non dev mode scenarios?

andytson-inviqa: Sorry I see from your comment above. I guess as we set do_development_start to only run if RUN_BUILD=true the previously more logical place was gone

andytson-inviqa:
I'd be open to some idea that can still run some stuff in dev mode if RUN_BUILD=false, but happy for this as it is, as it doesn't have negative impacts not in dev mode

PR #403, opened 2017-09-11
Split Magento 2 frontend build into install, run and clear cache.

kierenevans:
Support overriding bits of the frontend build rather than having to duplicate the whole thing.

### DIFF
--- a/magento2/usr/local/share/magento2/magento_functions.sh
+++ b/magento2/usr/local/share/magento2/magento_functions.sh
@@ -34,26 +34,42 @@ function do_magento_create_web_writable_directories() {
 }
 
 function do_magento_frontend_build() {
-  if [ -d "$FRONTEND_INSTALL_DIRECTORY" ]; then
-    mkdir -p pub/static/frontend/
+  do_magento_frontend_build_install
+  do_magento_frontend_build_run
+}
 
-  if [ -d "pub/static/frontend/" ] && [ "$IS_CHOWN_FORBIDDEN" != 'true' ]; then
-      chown -R "${CODE_OWNER}:${CODE_GROUP}" pub/static/frontend/
-    fi
-
-    if [ ! -d "$FRONTEND_INSTALL_DIRECTORY/node_modules" ]; then
-      as_code_owner "npm install" "$FRONTEND_INSTALL_DIRECTORY"
-    fi
-    if [ -z "$GULP_BUILD_THEME_NAME" ]; then
-      as_code_owner "gulp $FRONTEND_BUILD_ACTION" "$FRONTEND_BUILD_DIRECTORY"
-    else
-      as_code_owner "gulp $FRONTEND_BUILD_ACTION --theme='$GULP_BUILD_THEME_NAME'" "$FRONTEND_BUILD_DIRECTORY"
-    fi
-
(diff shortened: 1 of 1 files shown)


magento2/usr/local/share/magento2/magento_functions.sh
```diff
@@ -34,28 +34,44 @@ function do_magento_create_web_writable_directories() {
...
   fi
 }
 
+function do_magento_frontend_build() {
+  do_magento_frontend_build_install
+  do_magento_frontend_build_run
+}
+
+function do_magento_frontend_cache_clean() {
+  as_code_owner "npm cache clean --force"
```
andytson-inviqa: should we disable this in dev mode to keep things fast?

kierenevans: Good point - yes, it would allow a volume for /home/build/.cache/npm later on. I think we should do the same for the composer cache clear too.

magento2/usr/local/share/magento2/magento_functions.sh
```diff
@@ -34,28 +34,44 @@ function do_magento_create_web_writable_directories() {
...
 
-    if [ ! -d "$FRONTEND_INSTALL_DIRECTORY/node_modules" ]; then
-      as_code_owner "npm install" "$FRONTEND_INSTALL_DIRECTORY"
-    fi
-    if [ -z "$GULP_BUILD_THEME_NAME" ]; then
-      as_code_owner "gulp $FRONTEND_BUILD_ACTION" "$FRONTEND_BUILD_DIRECTORY"
-    else
-      as_code_owner "gulp $FRONTEND_BUILD_ACTION --theme='$GULP_BUILD_THEME_NAME'" "$FRONTEND_BUILD_DIRECTORY"
-    fi
+  as_code_owner "npm install" "$FRONTEND_INSTALL_DIRECTORY"
```
andytson-inviqa: Is npm standard in m2 or just the way we do things? should we check for the presence of package.json?

magento2/usr/local/share/magento2/magento_functions.sh
```diff
@@ -34,28 +34,44 @@ function do_magento_create_web_writable_directories() {
...
-    fi
+function do_magento_frontend_build_run() {
+  if [ ! -d "$FRONTEND_INSTALL_DIRECTORY" ]; then
+    return
+  fi
+
+  if [ -z "$GULP_BUILD_THEME_NAME" ]; then
+    as_code_owner "gulp $FRONTEND_BUILD_ACTION" "$FRONTEND_BUILD_DIRECTORY"
+  else
+    as_code_owner "gulp $FRONTEND_BUILD_ACTION --theme='$GULP_BUILD_THEME_NAME'" "$FRONTEND_BUILD_DIRECTORY"
```
andytson-inviqa: similar with gulp

magento2/usr/local/share/magento2/magento_functions.sh
kierenevans: Good point, I've made it check now

magento2/usr/local/share/magento2/magento_functions.sh
kierenevans: Good point, I've made it check now

PR #407, opened 2017-09-12
Add BATS Tests

kierenevans:
[BATS](https://github.com/sstephenson/bats) is a bash framework for testing.

I have added a few tests for some common functions in the ubuntu image with the aim to build this up over time.


sroze:
Nice :)

Issue #405, opened 2017-09-11
APP_ENDPOINT for symfony is /index.php instead of /app.php


TerjeBr:
PR #394  has broken our build!

/usr/local/share/40-stack has this line:
export APP_ENDPOINT=${APP_ENDPOINT:-/index.php}

Before DEFAULT_APP_ENDPOINT was computed and used to set APP_ENDPOINT to /app.php
in /usr/local/share/env/30-framework
This logic with DEFAULT_APP_ENDPOINT has been moved to /usr/local/share/env/45-framework, but that runs after /usr/local/share/40-stack, so the APP_ENDPOINT now always ends up as /index.php.

kierenevans:
Hi @TerjeBr, sorry that you encountered this.

Does using the stable tag instead of latest work for you?

Which symfony image variant are you using?

Do you have `symfony/flex` in your composer.lock anywhere?

TerjeBr:
Our Dockerfile begins with
`FROM quay.io/continuouspipe/symfony-php7.1-apache:latest`

We were using latest instead of stable because of some other issues. But may be we can go back to stable, I will have to check.

No, `symfony/flex` is not in our composer.lock anywhere.

TerjeBr:
This issue can now be closed, since #408 has been merged.

kierenevans:
Thanks @TerjeBr!

PR #408, opened 2017-09-12
Resolve directories earlier so Symfony endpoint change works

andytson-inviqa:


sroze:
This is to fix #405 by @TerjeBr, right?

andytson-inviqa:
@sroze yes, just testing it now

php/shared/usr/local/share/env/55-stack
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-export WEB_DIRECTORY=${WEB_DIRECTORY:-web}
```
TerjeBr: Why did you move `WEB_DIRECTORY` earlier? It seems kind of unrelated.

andytson-inviqa:
All looks ok now on Symfony 2.8/3.3(no flex) and 3.3(flex). endpoints resolve correctly, and loading the default page loads as expected for each

php/shared/usr/local/share/env/55-stack
andytson-inviqa: @TerjeBr it was because Flex moves the web directory to "public", the defaults when it was in 45-framework needed to be before the *-stack, so left 55-stack as the only option. Since 45-framework is now 37-framework and before 40-stack, it was able to be moved back to where it was before

PR #413, opened 2017-09-13
Add a `NON_PRIVELEGED_USER` environment variable to start services as the current user

sroze:


ubuntu/16.04/usr/local/share/bootstrap/common_functions.sh
```diff
@@ -39,9 +39,15 @@ get_user_home_directory() {
 }
 
 as_user() (
-  set +x
```
kierenevans: Can we reinstate this now that debugging is over, with a `set -x` on between lines 45 and 46

andytson-inviqa: I'd discussed this with @sroze before while we were pairing on it, but I agree now with @kierenevans

sroze: Done 👍

PR #416, opened 2017-09-14
Allow Spryker database scripts to run with a non-privileged PgSQL user

sroze:


spryker/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -49,10 +49,16 @@ do_spryker_app_permissions() {
 
 do_spryker_config_create() (
   set +x
-  echo "Creating Postgres Credentials file in /root/.pgpass"
+  if [ "$NON_PRIVELEGED_USER" = "true" ]; then
+    local target=~/.pgpass
```
andytson-inviqa: I suppose technically you can use ~/ for root too

spryker/usr/local/share/spryker/spryker_functions.sh
```diff
@@ -98,7 +104,7 @@ do_spryker_generate_files() {
 do_spryker_install() {
   # check if database exists (it is supposed to be created by postgres container)
   set +e
-    psql -U "$DATABASE_USER" -h "$DATABASE_HOST" -lqt | cut -d \| -f 1 | grep -q "$DATABASE_NAME"
+    psql -U "$DATABASE_USER" -h "$DATABASE_HOST" -lqt "$DATABASE_NAME" | cut -d \| -f 1 | grep -q "$DATABASE_NAME"
```
andytson-inviqa: I've not needed the database name for this before, I wonder why you do now

sroze: Because it's `postgres` by default, which is unauthorized for non privileged PostgreSQL users.

andytson-inviqa: perhaps something specific about the permissions the postgres docker image sets up

spryker/usr/local/share/spryker/spryker_functions.sh
sroze: You're right. Should I use `~` for all cases?

andytson-inviqa: it wouldn't hurt to

spryker/usr/local/share/spryker/spryker_functions.sh
andytson-inviqa: probably need to update the php/usr/share/assets/assets_functions.sh psql stuff too then

sroze: It did not cause any issue so far so I'd vote to keep them as they are

sroze: Actually, that makes sense to change them as well... doing that.

andytson-inviqa: what happens here about $DATABASE_PASSWORD? In the assets functions I declare PGPASSWORD just for the command

sroze: It should be in the `~/.pgpass` already, so no need of `PGPASSWORD`. `.pgpass` is the recommended way apparently.

sroze:
Are you okay with this @andytson-inviqa ?

andytson-inviqa:
I think we need to test this out in docker with a non-privileged user who's db is deleted

sroze:
Looks okay from @dantleech 👍

PR #419, opened 2017-09-20
Do not overwrite the `START_CRON` variable if already exists

sroze:


sroze:
I have to say... this still says this on Heroku:
```
$ cron -f
seteuid: Operation not permitted
```

andytson-inviqa:
https://devcenter.heroku.com/articles/scheduler

sroze:
@andytson-inviqa that's a quite succinct answer 😄 We have cronjobs to run every minute, so I don't think we should use the scheduler that (as far as I understood) is starting new dynos each time.

andytson-inviqa:
it goes on to mention clock processes and message workers https://devcenter.heroku.com/articles/scheduled-jobs-custom-clock-processes

andytson-inviqa:
though technically cronie is a clock process

PR #420, opened 2017-09-21
Split composer run into parts

kierenevans:
So we can ensure GITHUB_TOKEN is set in ~build/.composer/auth.json in the Magento2 image when called from `do_start` in https://github.com/continuouspipe/dockerfiles/blob/a1e76b97e70711cb1cbde4e4512d8e94768ffff3/magento2/usr/local/share/container/baseimage-30.sh#L16

It was observed that only the magento username/password were present in the composer auth file, so a manual run of `composer update <package>` in development was prompting for a token.


tkotosz:
thanks, looks good

PR #422, opened 2017-09-22
Transform URL services to other variables

sroze:


php/shared/usr/local/share/php/common_functions.sh
```diff
@@ -60,3 +60,33 @@ composer_package_compare() {
...
+  hostport=`echo $url | sed -e s,$userpass@,,g | cut -d/ -f1`
+  port=`echo $hostport | grep : | cut -d: -f2`
+  if [ -n "$port" ]; then
+    host=`echo $hostport | grep : | cut -d: -f1`
+  else
+    host=$hostport
+  fi
+
+  # extract the path (if any)
+  path="`echo $url | grep / | cut -d/ -f2-`"
```
andytson-inviqa: I'd still like to see the php parseurl used, just as it's one less thing that could be wrong.

Also Kieran is thinking about using some concurency, which could cause issues with these global proto/scheme/user/password/path will have issues with unless passed as a sub-shell

andytson-inviqa: not that I know a good way to use parseurl and collect the arguments though

sroze: OK, that's fair enough. Then... how would the result of this function look like then? /cc @kierenevans

sroze:
@kierenevans @dcole-inviqa @andytson-inviqa I know it's used by a few projects. Do you want to merge or take it over (i.e. do the change to use PHP instead of bash) ?

sroze:
I'll close it then.

PR #426, opened 2017-10-06
Fix latest tag not being used for master branch

kierenevans:
`expression` seem to win versus `value` now, causing `dev-master` tags to be used instead of `latest`
Please apply #427 first


sroze:
👍

PR #427, opened 2017-10-06
Fix elasticsearch build FROM_TAG.

kierenevans:
Fixes `elasticsearch:dev-master` not found. Please merge before #426 to avoid `elasticsearch:latest` being built from.


sroze:
👍

PR #430, opened 2017-10-10
Only install eZ DB if it doesn't already exist

maxbaldanza:


andytson-inviqa:
We have some code like this already that should probably be refactored for reuse. It additionally handles things like temporary network issues/credentials wrong not being interpreted as an empty db

maxbaldanza:
Happy for this to be closed if there's a better way of handling it 👍

ezplatform/usr/local/share/ez/ez_functions.sh
```diff
@@ -27,8 +27,30 @@ function do_ez_web_server_web_directory_non_writable() {
...
+     echo "false";
+   fi
+)
+
 function do_ez_install() {
-  do_symfony_console ezplatform:install "${EZPLATFORM_INSTALL_PROFILE}"
+  local DATABASE_EXISTS
+  DATABASE_EXISTS="$(check_ez_database_exists)"
+
+  if [ "$DATABASE_EXISTS" != "true" ]; then
```
sroze: Can't you use the `check_ez_database_exists` function directly in here?

ezplatform/usr/local/share/ez/ez_functions.sh
```diff
@@ -27,8 +27,30 @@ function do_ez_web_server_web_directory_non_writable() {
   fi
 }
 
+function check_ez_database_exists() (
+   set +e
```
sroze: Are your shell skills good enough to be able to extract the logic from [here](https://github.com/continuouspipe/dockerfiles/blob/395863e83ec2912c36062e4aa285a28e69440736/php/shared/usr/local/share/assets/assets_functions.sh#L96-L114) to a function that you could reuse in this example?

maxbaldanza: Thanks Sam 👍 I should be able to do that, just need to find some time to fit it in 😄

sroze:
@maxbaldanza any update on this one? :)

maxbaldanza:
Sorry @sroze had completely forgotten about this PR. 
I've just come back to it to try refactor the checking if the database exists as you suggested but it seems my bash skills are not what I thought. 
Would appreciate some help with this 😄

ezplatform/usr/local/share/ez/ez_functions.sh
```diff
@@ -27,8 +27,30 @@ function do_ez_web_server_web_directory_non_writable() {
...
+   local DATABASE_EXISTS
+   if [ -n "$DATABASE_PASSWORD" ]; then
+     mysql -h"$DATABASE_HOST" -u"$DATABASE_USER" -p"$DATABASE_PASSWORD" "$DATABASE_NAME" -e "SHOW TABLES; SELECT FOUND_ROWS() > 0;" | grep -q 1
+     DATABASE_EXISTS=$?
+   else
+     mysql -h"$DATABASE_HOST" -u"$DATABASE_USER" "$DATABASE_NAME" -e "SHOW TABLES; SELECT FOUND_ROWS() > 0;" | grep -q 1
+     DATABASE_EXISTS=$?
+   fi
```
andytson-inviqa: this is susceptible to network failures or invalid credentials being considered as an empty database. There is already code in the PHP image that does similar but without interpreting errors as empty database and could instead be extracted for reuse https://github.com/continuouspipe/dockerfiles/blob/master/php/shared/usr/local/share/assets/assets_functions.sh#L90

andytson-inviqa: sorry, sam already said the same latter stuff, just I didn't click the link to see beforehand

PR #438, opened 2017-10-26
Run the storage collectors before the search collectors

rgpjones:
* This should avoid am issue where the product data is missing
      in the storage, but is visible in the search results because
      the search collector has already run.


sroze:
Thank you @rgpjones

PR #440, opened 2017-10-27
Add RabbitMQ

dcole-inviqa:


rabbitmq/docker-compose.sample.yml
```diff
@@ -0,0 +1,14 @@
+version: '3'
+services:
+  rabbitmq:
+    image: "rabbitmq:3-management"
```
kierenevans: This would need to be `quay.io/continuouspipe/rabbitmq36-management:stable`

docker-compose.yml
```diff
@@ -559,6 +559,11 @@ services:
     depends_on:
       - ubuntu
 
+  rabbitmq36_management:
+    build:
+      context: ./rabbitmq/
+    image: quay.io/continuouspipe/rabbitmq36-management:latest
```
kierenevans: Is it possible to declare an external_rabbitmq36_management that this depends on too? Not sure we still use the build.sh script which pulls any external_* beforehand, but might be useful still.

kierenevans:
Looking good, we should get a README in here too :)

docker-compose.yml
dcole-inviqa: added external dependency

PR #442, opened 2017-11-15
Spryker: A failed frontend build should still fail the build

kierenevans:
Reverses the exit code ignoring change from https://github.com/continuouspipe/dockerfiles/commit/efac7db83edd722396c4838580470ca4bc00e18f

### DIFF
--- a/spryker/usr/local/share/spryker/spryker_functions.sh
+++ b/spryker/usr/local/share/spryker/spryker_functions.sh
@@ -84,10 +84,8 @@ spryker_build_assets() {
 }
 
 do_spryker_build_assets() {
-  set +e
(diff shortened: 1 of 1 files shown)


andytson-inviqa:
Why the original exit code ignoring change in the first place @kierenevans ?

andytson-inviqa:
and how does a frontend build fail if it's silently ignoring failures?

kierenevans:
I'm not entirely sure - I guess it was to let it build when we were testing it to get past the frontend build and onto "does the image work", but we shouldn't be ignoring errors anyway!

sroze:
Andy, happy with this one as well?

PR #446, opened 2017-12-13
Pass on the command to redis-cli

kierenevans:
Abstraction fail :D

### DIFF
--- a/redis/3-highly-available/README.md
+++ b/redis/3-highly-available/README.md
@@ -74,8 +74,7 @@ tasks:
                 cpu: 250m
                 memory: 500Mi
             command:
-              - container
-              - run_master_and_sentinel
+              - container run_master_and_sentinel
             environment_variables:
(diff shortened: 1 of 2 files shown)


sroze:
Haha. By the way, why `bash -c 'container ...'` in the `run` command? You shouldn't need it.

kierenevans:
Sorted 👍 
It was leftover from experimentation with commands, then it got too unwieldy and we moved to functions inside the docker image.

redis/3-highly-available/README.md
```diff
@@ -74,9 +74,8 @@ tasks:
                 cpu: 250m
                 memory: 500Mi
             command:
-              - /bin/bash
-              - -c
-              - 'container run_master_and_sentinel'
+              - container
```
sroze: you want that on one line tho

sroze: I mean `container run_master_and_sentinel` on one line

PR #450, opened 2018-01-08
Add support for redis password.

fratcliffe:


magento2/etc/confd/templates/magento/env.php.tmpl
```diff
@@ -40,6 +41,7 @@ return array (
                                     'server' => '{{ getenv "REDIS_HOST" }}',
 {{ end }}
                                     'port' => '{{ getenv "REDIS_PORT" }}',
+                                    'password' => '{{ getenv "REDIS_PASSWORD" }}',
```
sroze: If no `REDIS_PASSWORD` environment is present, this will be `password` as an empty string. Will it be ignored by the Redis driver? Or will it consider this as a password and the authentication will fail?

sroze: Maybe it'd be better to wrap it within a `if` if we don't know.

fratcliffe: 👍

PR #453, opened 2018-01-16
Use wget to download composer-setup.php and error if failing checksum match

andytson-inviqa:
PHP streams don't pick up HTTP_PROXY environment variable settings, and so don't try to run through a configured proxy.
Also the if condition wasn't erroring if the checksum didn't match

### DIFF
--- a/php/Dockerfile-apache
+++ b/php/Dockerfile-apache
@@ -46,13 +46,11 @@ RUN echo 'deb http://s3-eu-west-1.amazonaws.com/qafoo-profiler/packages debian m
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* \
  \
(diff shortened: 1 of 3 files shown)


sroze:
I guess this would apply for any other PHP scripts... Can't we configure PHP to use such proxy? (I google but found nothing, asking to be 100% sure 😉)

andytson-inviqa:
socket streams can do so using stream context settings I expect.

Composer respects the env vars (prob because of Guzzle doing so). This is specific to a forked source dockerfile builder requiring the proxy though, but not the actual app, I was just pushing the changes upstream.

Technically the Dockerfile needs ARGS HTTP... too to do so, but I figured to leave them out here and just make it easier should someone else need to do so as well.

andytson-inviqa:
can this be merged?

PR #456, opened 2018-01-26
Add HTTP2 toggling for Nginx and option for HTTP2 over HTTP

andytson-inviqa:
Nginx already had HTTP2 enabled

Latter is disabled by default as:
* Nginx doesn't support H2C, so is unable to negotiate HTTP version, so instead uses a different env var toggle
* H2C isn't supported by browsers, so has limited use

### DIFF
--- a/php/nginx/README.md
+++ b/php/nginx/README.md
@@ -115,8 +115,8 @@ WEB_HTTPS | Whether to support HTTPS traffic on the WEB_HTTPS_PORT | true/false 
 WEB_HTTPS_PORT | The port to serve the HTTPS traffic from | 0-65535 | 443
 WEB_HTTPS_OFFLOADED | Whether the HTTPS traffic has been forwarded without SSL to the HTTPS port | true/false | false
 WEB_HTTPS_ONLY      | Whether to redirect all HTTP traffic to HTTPS | true/false | $WEB_HTTPS (deprecated: if $WEB_HTTPS=true then false)
-WEB_HTTP2 | Whether to enable HTTP2 over TLS on HTTPS port. If WEB_HTTPS_OFFLOADED enabled then this is ignored due to Nginx not supporting h2c | true/false | true
-WEB_HTTP2_ONLY_HTTP | Whether to enable HTTP2 over plaintext on HTTP port (or HTTPS if WEB_HTTPS_OFFLOADED enabled). Nginx doesn't support h2c for plain HTTP protocol so will not support HTTP 1.1/1.0 if enabled | true/false | false
+WEB_HTTP2_TLS | Whether to enable HTTP2 over TLS on HTTPS port. If WEB_HTTPS_OFFLOADED enabled then this is ignored as TLS is not used | true/false | true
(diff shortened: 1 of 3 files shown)


andytson-inviqa:
need to double check Apache has mod_http2 enabled by default

andytson-inviqa:
Turns out Ubuntu hasn't packaged mod_http2, so I've removed the config for it

sroze:
There is no risks for existing websites to enable HTTP/2.0, right? Also, does this mean we cannot enable HTTP2 for HTTP & HTTPS?

andytson-inviqa:
We already have HTTP2 enabled for nginx image, so any risk was already there. Enabling HTTP2 on http port though is way too risky though as it will stop http 1.1 in Nginx.

php/nginx/README.md
```diff
@@ -115,6 +115,8 @@ WEB_HTTPS | Whether to support HTTPS traffic on the WEB_HTTPS_PORT | true/false
 WEB_HTTPS_PORT | The port to serve the HTTPS traffic from | 0-65535 | 443
 WEB_HTTPS_OFFLOADED | Whether the HTTPS traffic has been forwarded without SSL to the HTTPS port | true/false | false
 WEB_HTTPS_ONLY      | Whether to redirect all HTTP traffic to HTTPS | true/false | $WEB_HTTPS (deprecated: if $WEB_HTTPS=true then false)
+WEB_HTTP2 | Whether to enable HTTP2 over TLS on HTTPS port. If WEB_HTTPS_OFFLOADED enabled then this is ignored due to Nginx not supporting h2c | true/false | true
+WEB_HTTP2_ONLY_HTTP | Whether to enable HTTP2 over plaintext on HTTP port (or HTTPS if WEB_HTTPS_OFFLOADED enabled). Nginx doesn't support h2c for plain HTTP protocol so will not support HTTP 1.1/1.0 if enabled | true/false | false
```
sroze: Isn't it `WEB_HTTP2_ON_HTTP` ?

andytson-inviqa: I wanted to make it clear it would only do http2, not http 1.1, hence ONLY

andytson-inviqa: happy to go with a different variable name though, just still keeping it clear at the same time

sroze: `WEB_HTTP2_ONLY_ON_HTTP` ?

andytson-inviqa: `WEB_HTTPS_HTTP2`
`WEB_HTTP_HTTP2_ONLY`
?

andytson-inviqa: hmm, although that doesn't indicate that WEB_HTTPS in WEB_HTTPS_OFFLOADED enabled won't get WEB_HTTPS_HTTP2

sroze: mind = 💥 

So, we agree that this flag will configure if HTTP/2.0 is enabled on the non-secure HTTP port? 
If yes, what about:
1. `WEB_HTTP2_ON_HTTP_PORT`
2. `WEB_HTTP2_ON_NON_SECURE_HTTP`

IMHO, we don't care about the "only" as it's in the documentation (i.e. README file)

andytson-inviqa: `WEB_HTTP2_TLS`
`WEB_HTTP2_PLAINTEXT_NONBC`

PR #458, opened 2018-02-06
Support customising tideways hostname and environment

kierenevans:
Stops multiple hosts/environments appearing as `tideways-daemon (production)` in the Tideways UI which makes it hard to track down the site you are interested in.

### DIFF
--- a/tideways/usr/local/share/env/40-stack
+++ b/tideways/usr/local/share/env/40-stack
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+export TIDEWAYS_HOSTNAME=${TIDEWAYS_HOSTNAME:-tideways-daemon}
+export TIDEWAYS_ENVIRONMENT=${TIDEWAYS_ENVIRONMENT:-production}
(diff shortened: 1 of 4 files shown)


tideways/usr/local/share/env/40-stack
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+export TIDEWAYS_HOSTNAME=${TIDEWAYS_HOSTNAME:-tideways-daemon}
+export TIDEWAYS_ENVIRONMENT=${TIDEWAYS_ENVIRONMENT:-production}
```
sroze: Isn't it `default` by default? 🤔

tideways/usr/local/share/env/40-stack
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+export TIDEWAYS_HOSTNAME=${TIDEWAYS_HOSTNAME:-tideways-daemon}
```
andytson-inviqa: how about `WEB_HOST` as the default?

tideways/usr/local/share/env/40-stack
kierenevans: https://tideways.io/profiler/article/42-environments says `production`

tideways/usr/local/share/env/40-stack
kierenevans: WEB_HOST isn't normally on this container, but we could add it

andytson-inviqa: ah right, nvm, I wasn't thinking it was a separate image

PR #460, opened 2018-02-09
Add PHP 7.2 images in base and Symfony varients

andytson-inviqa:
PHP 7.2 removed mcrypt extension and moved to pecl.php.net.

Magento and Drupal aren't yet ready for it. Both additionally require mcrypt, however it'd be bad practice to include the pecl one. Instead a mcrypt polyfill could be used by the app.

### DIFF
--- a/continuous-pipe.yml
+++ b/continuous-pipe.yml
@@ -58,6 +58,13 @@ tasks:
           image: quay.io/continuouspipe/nginx
           tag: ${FROM_TAG}
           reuse: false
+        php72_apache:
+          image: quay.io/continuouspipe/php7.2-apache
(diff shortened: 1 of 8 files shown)


kierenevans:
Hey @andytson-inviqa! Great work so far - thank you.
There's a missing`symfony_php72_apache` target in `continuous-pipe.yml`.
Having a look at the :latest tag issue that failed the CP build.

kierenevans:
https://github.com/continuouspipe/dockerfiles/pull/461 should sort the FROM_TAG issue out 🤞

sroze:
We need to regenerate the `stable` tags though.

PR #462, opened 2018-02-12
Fix Opcache's validate_timestamps setting always being 1

kierenevans:
The check against DEVELOPMENT_MODE not being "true" did succeed, but the PHP_OPCACHE_VALIDATE_TIMESTAMPS value was already set, so it got used in 55-stack


sroze:
Why is that for? When you set the environment variable `PHP_OPCACHE_VALIDATE_TIMESTAMPS=0` yourself, it disables it on development mode? 🤔

kierenevans:
You could set to `0` yourself which would make sure no changes to php files are reflected in the loading of the next page, but you wouldn't want that in development.
I think the original intent of the setting was to ensure this is set to 0 unless DEVELOPMENT_MODE is set to "true", as it gives better performance - it's just it got set to 1 by mistake as it inherited the first setting of it.

kierenevans:
Sorry, re-read your message - setting it to `0` yourself in development will persist correctly.

kierenevans:
You're right, this makes no sense :joy:

PR #464, opened 2018-02-15
TLS cipher improvements

andytson-inviqa:
Due to ROBOT and SWEET32 security flaws, some ciphers need to be able to be removed. They are ok to be removed for most user browsers, but may impact corporate browsers and api clients, so can't be disabled without reviewing that.

Additionally it appears from the cipher list order that ECDSA certificate ciphers are less preferred than RSA certificate ciphers (aRSA before ECDSA in alphabet?). Apache/nginx support both certificates being loaded (although not our configuration by itself), so these should be ordered based on security improvement preference.


andytson:
I'm going to rename ROBOT to RSA something. It's not a fix for ROBOT (which is a implementation issue rather than cipher), but a best practice given flaws like this are happening.

andytson:
I'm going to rename ROBOT to something else, as openssl isn't vulnerable to the specific attack. Instead RSA encryption is just considered weak to future attacks of similar nature.

PR #465, opened 2018-02-16
dhparam generation

andytson-inviqa:
Apache/Nginx by default (through openssl defaults) use a weak 1024 bit dhparam generated at startup and only regenerated on reload or a new container.

Generating or loading an external dhparam file allows for this to be made a stronger 2048 bit.

Additionally scheduling regular daily dhparam regeneration ensures security.

This feature is off by default, and turned on using WEB_SSL_DHPARAM_ENABLE environment variable.


nginx/etc/confd/templates/supervisor/dhparam_regenerate.conf.tmpl
```diff
@@ -0,0 +1,5 @@
+[program:dhparam-regenerate]
+command = supervisorctl --user supervisor --password supervisor start dhparam-regenerate
```
kierenevans: Will this be an infinite loop?

ubuntu/16.04/usr/local/share/bootstrap/common_functions.sh
```diff
@@ -291,6 +291,13 @@ function test_remote_ports() {
   done
 }
 
+function supervisor_signal() {
+    local -r SIGNAL=$1
+    local -r PROGRAM_NAME=$2
+
+    supervisorctl --username supervisor --password supervisor signal "$SIGNAL" "$PROGRAM_NAME"
```
kierenevans: Awesome!

php/shared/etc/confd/templates/supervisor/dhparam_regenerate.conf.tmpl
```diff
@@ -0,0 +1,5 @@
+[program:dhparam-regenerate]
+command = supervisorctl --user supervisor --password supervisor start dhparam-regenerate
```
kierenevans: Is this an infinite loop?

nginx/etc/confd/templates/supervisor/dhparam_regenerate.conf.tmpl
andytson-inviqa: no autorestart and startsecs set to 0, so it wont restart under any condition.

andytson-inviqa: oh oops I see what you mean

andytson-inviqa: I tested it somehow, maybe before I "corrected" a non-issue in `container --silent dhparam_regenerate`

andytson-inviqa: fixed

php/shared/etc/confd/templates/supervisor/dhparam_regenerate.conf.tmpl
andytson-inviqa: fixed

sroze:
Oups, conflicts :)

andytson-inviqa:
@sroze fixed, just was one of those lines changed immediately after a changed line conflicts that could really auto-resolve but for some reason doesn't

PR #466, opened 2018-02-23
Fix n98 download if the bin directory doesn't get created

kierenevans:
(e.g. if composer is run with --no-dev).


magento1/usr/local/share/magento1/magento_functions.sh
```diff
@@ -2,6 +2,7 @@
 
 function do_magento_n98_download() {
   if [ ! -f bin/n98-magerun.phar ]; then
+    as_code_owner "mkdir -p bin"
```
andytson-inviqa: for a second I was wondering why `-p` when if you're cd'd into /app then /app obv exists, but remembered it also silences the already exists errors :)

PR #467, opened 2018-02-28
MISC Add linux compatible permission update.

jacs-inviqa:
The uid and gid of the 'build' user should match the host users uid/gid

### DIFF
--- a/ubuntu/16.04/usr/local/share/bootstrap/update_permissions.sh
+++ b/ubuntu/16.04/usr/local/share/bootstrap/update_permissions.sh
@@ -24,14 +24,20 @@ function update_permissions() {
         CODE_OWNER="$APP_USER"
         export APP_USER
         export CODE_OWNER
-
         if [[ "$group" = UNKNOWN ]]; then
             APP_GROUP="$APP_USER"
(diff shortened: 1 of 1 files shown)


ubuntu/16.04/usr/local/share/bootstrap/update_permissions.sh
```diff
@@ -24,14 +24,20 @@ function update_permissions() {
...
         if [[ "$group" = UNKNOWN ]]; then
             APP_GROUP="$APP_USER"
             CODE_GROUP="$APP_GROUP"
             export APP_GROUP
             export CODE_GROUP
             groupadd --system --gid "$group_id" "$APP_GROUP" || groupmod -g "$group_id" "$APP_GROUP"
+        else
+            APP_GROUP="$group"
+            CODE_GROUP="$APP_GROUP"
+            export APP_GROUP
```
sroze: There’s some duplication here that prI à my could have needed a small refactoring :)

sroze: Can’t update on mobile... but « prl à my » = « probably » 😅

PR #468, opened 2018-02-28
Add check for a custom plan sh override file

jacs-inviqa:
Sometimes it is necessary to add local changes required only for the dev machine, even if these are sometimes temporary...


sroze:
I like the idea. Could you add this to the documentation as well?

sroze:
ping @jacs-inviqa

kierenevans:
Documentation added. If it's okay, let's merge this!

PR #469, opened 2018-03-08
Build in groups to avoid timeout

kierenevans:
History of builds is not great -  some images haven't been built in ages due to timeouts.

I have a hunch it's per a TideCommand timeout so splitting into two groups should hopefully help us get past it.

Tagging as "Ready for Review" to see how far the build gets.


kierenevans:
Yep!

PR #473, opened 2018-04-09
Convert mongodb CMD wrapper to upstream entrypoint script process

andytson-inviqa:
This has most backwards-compatibility in place whilst deprecating environment variables.

There are a few minor backwards-compatibility breaks:

* Users are only created on first initialisation of the mongo database. Previously they'd be updated on every restart
* Generating an admin/root user is the only way to enable authentication. Previously users without admin privileges could be only those needed. This is miitigated to an extent that MONGODB_AUTH_ENABLED does this for you, which was required to enable authentication before (it isn't needed for that now)

Additionally as 2.* is now all EOL, and the upstream mongo:2.6 isn't being updated, 2.6 has been removed here as well


andytson-inviqa:
I recently had to create new users for an existing mongodb, and didn't want to go through the process of either manually assigning them or recreating the data volume, so I'm not keen on this BC break and would like a solution that doesn't do so, so I'm closing this.

PR #474, opened 2018-04-13
Add a comment to build.sh on why the external_ images are pulled first

andytson-inviqa:
rather than docker-compose build --pull used


sroze:
Thank you :)

PR #476, opened 2018-04-19
Downgrade php redis extension to 3.1.6 to fix backwards incompatibility

andytson-inviqa:
Version 4 changed the function api and broke some wrappers (perhaps those wrappers were too tightly coupled as well)


andytson-inviqa:
I only know of Snc\RedisBundle atm tho few people are running `docker-compose build --pull` on their machines (also why it was only found out 10 days after).
The extension maintainers haven't also made it easy to find what the backwards-incompatibilities are.

andytson-inviqa:
I'd suggest we consider some limited version bumping in the future to wipe BC workarounds occasionaly

PR #481, opened 2018-05-14
remove duplicate pgsql php extension from nginx/apache

Oliboy50:
It's not a big deal, I don't think it will reduce the build time... but still :)

ℹ️  It seems to have been introduced in https://github.com/continuouspipe/dockerfiles/pull/331


kierenevans:
Thank you @Oliboy50!

PR #486, opened 2018-07-18
Fix typo in postfix config

kierenevans:


php/shared/etc/confd/templates/postfix/main.cf.tmpl
```diff
@@ -6,7 +6,7 @@ inet_protocols = ipv4
 mydestination = {{ getenv "WEB_HOST" }}, localhost.localdomain, localhost
 mydomain = {{ getenv "WEB_HOST" }}
 myhostname = {{ getenv "WEB_HOST" }}
-myorgin = $myhostname
+myorigin = $myhostname
```
andytson-inviqa: ironically this line wasn't needed (nor mydomain), as the default is correct too

PR #503, opened 2018-10-04
TCP-2606: Magento2 add test to clear redis chache without using FLUSHDB

marcomc:
DO NOT MERGE YET, it includes test commends

### DIFF
--- a/magento2/usr/local/share/magento2/magento_functions.sh
+++ b/magento2/usr/local/share/magento2/magento_functions.sh
@@ -159,7 +159,8 @@ function do_magento_assets_download() {
   fi
 }
 
-function do_magento_clear_redis_cache() {
+function do_magento_clear_redis_cache() (
+  set -o pipefail
   if [ "$MAGENTO_USE_REDIS" != "true" ]; then
     return
   fi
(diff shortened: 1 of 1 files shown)


magento2/usr/local/share/magento2/magento_functions.sh
```diff
@@ -176,8 +176,16 @@ function do_magento_clear_redis_cache() {
...
+  #redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_CACHE_DATABASE" "FLUSHDB"
+  #redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_FULL_PAGE_CACHE_DATABASE" "FLUSHDB"
+
+  # This is just a test to see the output of --scan
+  # to be deleted after making the tests
+  redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT"   --scan | xargs
+
+  #this command should delete the cache without blocking the database
+  redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_CACHE_DATABASE" --scan | xargs redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_CACHE_DATABASE" del
+  redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_FULL_PAGE_CACHE_DATABASE" --scan | xargs redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_CACHE_DATABASE" del
```
kierenevans: In the `del` command here, `$MAGENTO_REDIS_CACHE_DATABASE` is used when it should be `$MAGENTO_REDIS_FULL_PAGE_CACHE_DATABASE` to match the --scan command.

magento2/usr/local/share/magento2/magento_functions.sh
```diff
@@ -176,8 +176,16 @@ function do_magento_clear_redis_cache() {
...
-  redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_FULL_PAGE_CACHE_DATABASE" "FLUSHDB"
+  #redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_CACHE_DATABASE" "FLUSHDB"
+  #redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_FULL_PAGE_CACHE_DATABASE" "FLUSHDB"
+
+  # This is just a test to see the output of --scan
+  # to be deleted after making the tests
+  redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT"   --scan | xargs
+
+  #this command should delete the cache without blocking the database
+  redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_CACHE_DATABASE" --scan | xargs redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -n "$MAGENTO_REDIS_CACHE_DATABASE" del
```
kierenevans: xargs will try to put up to the maximum number of characters for arguments on a single line when executing based on the environment limits.
For example:
```
root@cdc6260bf875:/app# xargs --show-limits
Your environment variables take up 9243 bytes
POSIX upper limit on argument length (this system): 2085861
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 2076618
Size of command buffer we are actually using: 131072
Maximum parallelism (--max-procs must be no greater): 2147483647
```

If lots of entries are present in the redis database, it will try to clear as many of them as possible at once, which might mean that we are in a no better situation than using `flushdb`.

We can send arguments to xargs to control this behaviour though, for example`xargs --max-args=100` will clear a maximum 100 keys at once.

When testing with an empty database, redis-cli complained if there were no keys to delete, so adding `--no-run-if-empty` to `xargs` will help with that.

PR #520, opened 2019-04-18
Remove debian:wheezy based builds as it's officially EOL

kierenevans:
https://wiki.debian.org/DebianWheezy states that Debian Wheezy has been EOL since 2018-05-31.
It has recently been removed from the debian mirrors (as announced here: https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html ).

During a review of the external images we use for #519 it was found that the following images are based on debian:wheezy:
* mongo:2
* gcr.io/google_containers/redis:v1

In the case of mongo, 2.6 was EOL in October 2016 according to https://www.mongodb.com/support-policy and https://hub.docker.com/_/mongo no longer lists 2.x as a supported docker image version, so no updates are possible.

In the case of highly available redis, a newer image from 2017 is listed, however it's based on Ubuntu 14.04 which is also EOL:
https://console.cloud.google.com/gcr/images/google-containers/GLOBAL/redis?gcrImageListsize=30


sroze:
Very nice housekeeping 💚

PR #523, opened 2019-07-17
Fix the Spryker image build for PHP 7.3

unfunco:
I missed a couple of things :man_facepalming: Hopefully this fixes it.

### DIFF
--- a/spryker/README.md
+++ b/spryker/README.md
@@ -2,7 +2,7 @@
 
 In a Dockerfile for PHP and NGINX:
 ```Dockerfile
-ARG PHP_VERSION=7.3 # 7.1 and 7.2 are also available.
+ARG PHP_IMAGE_VERSION=7.3 # 7.1 and 7.2 are also available.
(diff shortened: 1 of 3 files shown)


spryker/README.md
```diff
@@ -2,7 +2,7 @@
 
 In a Dockerfile for PHP and NGINX:
 ```Dockerfile
-ARG PHP_VERSION=7.3 # 7.1 and 7.2 are also available.
+ARG PHP_IMAGE_VERSION=7.3 # 7.1 and 7.2 are also available.
 FROM quay.io/continuouspipe/spryker-php${PHP_VERSION}-nginx:stable
```
kierenevans: The `ARG` above only applies to this `FROM`, so needs to be used as PHP_IMAGE_VERSION or ARG adjusted to PHP_VERSION

unfunco: :man_facepalming: Sorry, missed that bit too, yep, I renamed so it was consistent with the build arg naming, updated.

